PR #134, opened 2014-08-11
Implements backtracking cut operator.

brandonbloom:
Speculative pull-request. Please consider this a feature suggestion / idea thread until further notice. Code seems to work, but not carefully vetted yet.

Circle CI uses schema for validation of config files. Some uses of `either` in Circle's schemas cause crummy error messages. For example:
screen shot 2014-05-21 at 2 06 25 pm

In this case, we know we've got a Heroku config and not a command line, but we get a more general error because both branches of an `(either String Heroku)` node fail. Deeper in the schema, a sensible error message was produced, but backtracking masked it.

Here's a simpler test case:

``` clojure
(require '[schema.core :as s] :reload-all)

(def CommandMap {:command s/Str :args [s/Str]})
(def Command (s/either s/Str CommandMap))

(s/validate Command "echo hello there") ; success
(s/validate Command {:command "echo" :args ["hello" "there"]}) ; success
(s/validate Command {:command "echo" :args ["hello" :there]}) ; fails with bad error
```

The low quality error is:

```
Value does not match schema: (not (some (check % a-clojure.lang.PersistentArrayMap) schemas))
```

This pull request implements an "impure" `cut` operator in the style of Prolog. When used carefully, cuts can substantially cull the search space, resulting in better performance and, importantly, better error messages. Continuing with the same example:

``` clojure
(def Command (s/either s/Str (s/cut CommandMap))) ; note the cut

(s/validate Command {:command "echo" :args ["hello" :there]}) ; fails with *good* error
```

The new error is:

```
Value does not match schema: {:args [nil (not (instance? java.lang.String :there))]}
```

This can work more generally across a schema AST, since it's implemented via exception handling. A cut can be placed in a `both` or similar to guide cutting. For example:

``` clojure
;; Guarded cut
(s/both (s/pred map?) (s/cut CommandMap))
```

It may actually make sense to bake guarding support in to the `cut` function, since the tagged-node single-arity style usage is a bit awkward. I just wanted to get this PR up for discussion before getting in to the details of what API would be best, if this feature is decided to be a good idea at all.

Of course, like Prolog's cut, unsoundness can creep in to validation if cut is used improperly. It's a dangerous, but useful side-effecting operation.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -423,6 +423,25 @@
   (NamedSchema. schema name))
 
 
+;;; cut (eliminate backtracking up to the most recent choice point)
+
+(defrecord Cut [schema]
+  Schema
+  (walker [this]
+    (let [sub-walker (subschema-walker schema)]
+      (fn [x]
+        (throw (ex-info "cut" {::walker sub-walker})))))
+  (explain [this]
+    (list 'cut (explain schema))))
+
+(defn cut
+  "Cuts other branches from this subschema to the nearest choice point.
+  Improper use can create unsound schemas, but often useful for eliminating
+  slow backtracking and improving validation error messages."
+  [schema]
+  (Cut. schema))
+
+
 ;;; either (satisfies this schema or that one)
 
 (defrecord Either [schemas]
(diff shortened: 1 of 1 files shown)


brandonbloom:
Watching @cursiveclojure's Conj talk about PEG-parsing macros reminded me of this ticket. Colin points out a great heuristic for dealing with backtracking: Keep track of the "deepest match" as you go, then report that error message instead of the error message from the decision point. Bummed I didn't think of that!

w01fe:
Ah, that's interesting -- thanks!  Wish I had thought of it too :).  

In any case, we've already gone the route of deprecating `either` in the 1.0 release, so the (built-in) schema types are now all backtrack-free.

cursiveclojure:
I haven't had a chance to read all this yet, but I will do since I'm planning to implement `cut` in my parser too, although for other reasons.

The furthest-match heuristic was originally proposed by Ford, but the best reference IMO is here: [Error Reporting in Parsing Expression Grammars](http://arxiv.org/abs/1405.6646) which has the semantics fully defined. The furthest matches are combined in order to produce messages like "Expected string, vector or keyword but found *whatever*", or even "Expected destructuring, but found *whatever*".

I'm planning to implement `cut` so that I can report multiple errors per form. This is useful because I'll be using the partial parse results to implement editor functionality like completions that are specific to particular code sites. Currently if there's an error near the beginning of a large form (i.e. `ns` or `defrecord`) I can still get partial results up to that point, but by using `cut` to bound the scope of the error I'll be able to continue parsing and hopefully match the rest of the form and get a larger partial result.

brandonbloom:
Ah, interesting! That problem doesn't really come up often in schema validation. It doesn't usually arise for associative structures or homogenous sequences. Mainly arrises when you want to parse some sequence where the correctness of item N depends on items [0..N), which is true of many macros' grammars.

PR #167, opened 2014-11-03
implementation of delimited sequence schema

davegolland:
Implements delimited sequence schema for matching delimited sequences.

An example of a delimited sequence is: `(s/delimited [s/Int] :delimiter [s/Str])` which would match `[1 2 3 4 :delimiter "a" "b" "c"]`. Here there is a sequence of integers, a single delimiter, and then a sequence of strings all concatted into the same vector. The matching is implemented by splitting the sequence on the delimiter, and then recursively matching each side. In the event of multiple delimiters, there is a flag to determine whether to match on the first or last occurrence.

The motivation is to make schema rich enough to express the sequences used in clojure's binding forms, such as: `(let  [[a b :as args] [1 2]] ...)`. In this binding, the `:as` is the delimiter, and the left schema is a sequence of symbols, the right schema is a single symbol `[(s/one clojure.lang.Symbol "binding name")]`.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -858,6 +858,50 @@
    (one second-schema second-name)])
 
 
+;; A delimited sequence schema matches sequences where the first part
+;; of the sequence matches the left-schema, and the last part of the
+;; sequence matches the right-schema. The transition point is
+;; indicated by the location of the delimiter.  All matched sequences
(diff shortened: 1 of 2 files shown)


w01fe:
Closing for now, let's revisit once the new protocol is in place.  Although it would be interesting to think about how the new protocol makes this and more parsing-like applications easier or harder.

PR #177, opened 2015-01-31
Support java.util.List in vector schemas

loganlinn:
Classes like java.util.ArrayList do not currently pass [] schemas because
they don't pass `sequential?`. However, since they are seqable, we are able
to walk the schema as one would expect. Expands the conditional to accept
instances of java.util.List.


w01fe:
:+1:

PR #190, opened 2015-03-07
Allow use of schema.core without :includes-macros in Cljs

Deraen:
Hi,

From version 2755 ClojureScript has supported loading macros automatically if the required ns (e.g. `schema.core`) loads it's own macros using `:require-macros`.

However, there is a small problem. Upgrading the ClojureScript dependency breaks a test: https://github.com/Prismatic/schema/blob/master/test/cljx/schema/core_test.cljx#L984

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -5,7 +5,7 @@
             :url "http://www.eclipse.org/legal/epl-v10.html"}
 
   :profiles {:dev {:dependencies [[org.clojure/clojure "1.6.0"]
-                                  [org.clojure/clojurescript "0.0-2665"]
+                                  [org.clojure/clojurescript "0.0-2760"]
                                   [org.clojure/tools.nrepl "0.2.5"]]
                    :plugins [[com.keminglabs/cljx "0.6.0" :exclusions [org.clojure/clojure]]
                              [codox "0.8.8"]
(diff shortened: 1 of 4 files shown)


w01fe:
Thanks for the PR!  This will definitely be nice in the future, but for the time being I don't think we want to break compatibility with older versions of Clojurescript -- so adding the `:require-macros` to schema.core is probably okay, but removing the `:include-macros true` from the other namespaces should probably wait.  Sound right?    
 
What's breaking with the test -- does it look like a real issue or just a test problem?  

Thanks again.

Deraen:
Right, adding :require-macros shouldn't break compatibility with older versions. I can add :include-macros back to schema.coerce and schema.test. How about the tests? Currently 2120 is supported but I think tests are only run using 2665, so is upgrading the tests to use 2760 okay?

The test problem is just a test problem. The validation throws a exception but it looks like converting the exception to string fails. It can be fixed by using message method of js/Error instead of str.

Deraen:
Str not working with ExceptionInfo is definitely a Cljs problem, I'll see that it will be fixed there.

Deraen:
Ah, it has already been fixed and it should work with the next Cljs release. http://dev.clojure.org/jira/browse/CLJS-1064

w01fe:
Awesome, thanks for the update and the details on the bug -- sounds like we can just let it be for the time being then.

PR #196, opened 2015-03-18
Update README.md

kernelp4nic:
Hello! First of all, thank you guys for this amazing library :)

I'm adding an example of s/conditional to the Readme,
I was struggling with this and I found the feature inside core.cljx but nothing outside to point me out
to the right place.

Let me know if it's inside the right section or if something needs to be changed.


w01fe:
Thanks for your contribution and the kind words! 

This example is helpful, but I think it's a bit long for the main README.  I think a simpler example illustrating `s/conditional` could go well in the "Other Schema Types" section (e.g. the one in the tests), and if you still think this longer example adds something, we could maybe start a new file for more detailed documentation of this kind?

kernelp4nic:
Neat, I will make the example more simple, I think just mentioning this capability adds value for the newcomers.

w01fe:
Closing for inactivity; please reopen if you get to the update.  Thanks!

PR #199, opened 2015-04-08
Set up Travis CI

w01fe:


hogaur:
We shouldn't be needing this. This issue is stale in my opinion. Also, given that one needs credits to use the Travis CI with [this](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) recent change, we are better off without it.

w01fe:
Yes, this was my old attempt before we set up CircleCI.

PR #201, opened 2015-04-09
improve strict-map error message

mrrodriguez:
It appears that the error message thrown from `strict-map->*` factory function generated via schema.core/defrecord has the "expected"/"got" parts flipped with one another.

e.g.
```clj

(schema.core/defrecord MyRec [x y])

(strict-map->MyRec {:x 1 :y 2 :z 3})
;; Throws:
;; RuntimeException Wrong number of keys: expected (:x :y :z), got (:x :y)  user/strict-map->MyRec (form-init8889428310851646570.clj:1)

;; Should be:
;; RuntimeException Wrong number of keys: expected (:x :y), got (:x :y :z)  user/strict-map->MyRec (form-init8889428310851646570.clj:1)

(strict-map->MyRec {:z 3})
;; Throws
;; RuntimeException Wrong number of keys: expected (:z), got (:x :y)  user/strict-map->MyRec (form-init8889428310851646570.clj:1)

;; Should be:
;; RuntimeException Wrong number of keys: expected (:x :y), got (:z)  user/strict-map->MyRec (form-init8889428310851646570.clj:1)


```

Interestingly when the correct number of keys is found, the code goes down a different path and throws a different error message.  I'm not sure this is intentional.

e.g.
```clj
(strict-map->MyRec {:x 1 :z 3})
;; Throws:
;; RuntimeException Key :y not found in {:z 3, :x 1}  user/strict-map->MyRec (form-init8889428310851646570.clj:1)

```

The validation for key count passes.  This exception is actually from `schema.macros/safe-get`.
However, this error message is technically correct so I haven't addressed it here.  It may be a little less descriptive than
the incorrect count messages though.

### DIFF
--- a/src/clj/schema/macros.clj
+++ b/src/clj/schema/macros.clj
@@ -330,7 +330,7 @@
              [~map-sym & [drop-extra-keys?#]]
              (when-not (or drop-extra-keys?# (= (count ~map-sym) ~(count field-schema)))
                (error! (utils/format* "Wrong number of keys: expected %s, got %s"
-                                      (sort (keys ~map-sym)) (sort ~(mapv keyword field-schema)))))
+                                      (sort ~(mapv keyword field-schema)) (sort (keys ~map-sym)))))
              (new ~(symbol (str name))
                   ~@(map (fn [s] `(safe-get ~map-sym ~(keyword s))) field-schema)))))))
 


w01fe:
Good catch -- thanks for your contribution!  

It's structured the way it is for performance, but there's no reason the error messages couldn't look the same / use the same code, this was just the simplest route given the overall structure.  Further improvements welcome :).

PR #207, opened 2015-04-21
Add recursive schemas for ClojureScript

viesti:
Relies on var support that was added in 0.0-2850

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -544,54 +544,38 @@
 ;; Clojure and ClojureScript vars.
 
 #+clj
-(do
-  (clojure.core/defn var-name [v]
-    (let [{:keys [ns name]} (meta v)]
-      (symbol (str (ns-name ns) "/" name))))
+(clojure.core/defn var-name [v]
+  (let [{:keys [ns name]} (meta v)]
+    (symbol (str (ns-name ns) "/" name))))
 
-  (clojure.core/defrecord Recursive [derefable]
-    Schema
-    (walker [this]
-            (let [a (atom nil)]
-              (reset! a (start-walker
-                         (let [old subschema-walker]
-                           (clojure.core/fn [s] (if (= s this) #(@a %) (old s))))
-                         @derefable))))
-    (explain [this]
-             (list 'recursive
-                   (if (var? derefable)
-                     (list 'var (var-name derefable))
-                     (format "%s@%x"
-                             (.getName (class derefable))
-                             (System/identityHashCode derefable))))))
+(#+clj clojure.core/defrecord #+cljs cljs.core/defrecord Recursive [derefable]
+  Schema
+  (walker [this]
+          (let [a (atom nil)]
+            (reset! a (start-walker
+                       (let [old subschema-walker]
+                         (clojure.core/fn [s] (if (= s this) #(@a %) (old s))))
+                       @derefable))))
+  (explain [this]
+           #+clj
+           (list 'recursive
+                 (if (var? derefable)
+                   (list 'var (var-name derefable))
+                   (format "%s@%x"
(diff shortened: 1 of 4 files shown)


src/cljx/schema/core.cljx
```diff
@@ -573,6 +573,26 @@
...
+    Schema
+    (walker [this]
+      (let [a (atom nil)]
+        (reset! a (start-walker
+                   (let [old subschema-walker]
+                     (clojure.core/fn [s] (if (= s this) #(@a %) (old s))))
+                   @derefable))))
+    (explain [this]
+      (let [{:keys [ns name]} (meta derefable)]
+        (list 'recursive (str ns "/" name)))))
```
w01fe: can we share more of this code with the Clojure version?

src/cljx/schema/core.cljx
```diff
@@ -573,6 +573,26 @@
...
+
+  (cljs.core/defn recursive
+  "Support for (mutually) recursive schemas by passing a var that points to a schema,
+  e.g (recursive #'ExampleRecursiveSchema)."
+  [schema]
+  (Recursive. schema)))
```
w01fe: What will happen on earlier versions of cljs?  Ideally we don't break things for earlier versions.  

Can you please also add an entry to the changelog describing the addition and the new minimum cljs version?

test/cljx/schema/core_test.cljx
```diff
@@ -263,59 +263,58 @@
...
+             :children [{:node-id 1
+                         :children [{:node-id 4}]}
+                        {:node-id "invalid-id"}]})
+
+  (valid! TestBlackNode {:black {}})
+  (valid! TestBlackNode {:black {:red {:black {}}}})
+  (invalid! TestBlackNode {:black {:black {}}})
+  (invalid! TestBlackNode {:black {:red {}}})
+
+  #+clj(let [rec (promise)
```
w01fe: Can you put a newline before the open paren please (and below)?

w01fe:
Looks good modulo comments, thanks for your contribution!

src/cljx/schema/core.cljx
viesti: Actually, now that I look a bit closer, static vars got introduced in 0.0-2496 so this should actually work with 0.0-2760. At least ```lein test``` passes.

src/cljx/schema/core.cljx
viesti: Good observation! I was focused on just making it work at first :). Will try to grab time to look at code sharing.

test/cljx/schema/core_test.cljx
viesti: Will do, thanks!

w01fe:
Thanks!  LMK when you have a chance to update with the code sharing.  Or if you don't get a chance in a little while, I might take a crack at it.

viesti:
If you have brain cycles to spare for code sharing, then go for it! I've a bit busy atm. :)

w01fe:
Not many spare cycles but I'll get to it when I can -- if you get a chance
before then, just bump the thread to let me know and I'll get it
merged/released ASAP.  Thanks!

viesti:
Spent the cycles from this evening :) Now the main part is shared and platform specific code is separated in a more fine grained way.

Changed also the test to use ```atom``` instead of ```promise``` to be able to share more of the test code.

w01fe:
Thanks!  Closing in favor of #208.  I ended up squashing your changes into a single commit and making a couple small tweaks, please let me know if that's okay and I'll go ahead and merge (I preserved your authorship but probably didn't do it the "right" way).

Thanks again for your contribution!

viesti:
Squash is ok, ended up working on many intermediate steps. Thanks! :)

PR #211, opened 2015-04-28
defschema with :full-name meta-data

### LINKED ISSUE #210: more distinct names for Schemas
hi,

Would be great if schemas would have more distinct names. Currently one can create multiple schemas into separate namespaces with same name but with different value. If schema name would contain the namespace also, so they would be unique and by so, more useful for documentation purposes.

Ideas:
1) change the current `:name` to contain the namespace-part too (a breaking change, bad)
2) add more meta-data to schemas, `:full-name` with the ns + the file-name & line-number where it was defined. Would help a lot on the systems that use Schema for documentation.

would be happy to create a PR for the 2 if that is a valid approach.

https://github.com/frankiesardo/pedestal-swagger/issues/7

https://github.com/metosin/ring-swagger/blob/6c0ada07e146a3247f58f92347d509e8f9d959d6/src/ring/swagger/core.clj#L104-L109

ikitommi:
`defschema` now adds `:full-name` metadata with fully-qualified schema-name. One can retrieve the full-name with helper function `full-schema-name`. Resolves #210.

This also creates two nice bonuses:

1) one can resolve the `:full-name` and get the Schema Var (in the Clojure-side only) and fetch all the extra meta-data out that: docstring, line-number, file etc. Great for documentation. See `schema.core-test.cljx` for details.

2) one **could** create tooling with this to check whether Schema definitions and linked Schemas are in sync:

```clojure

;; defina a schema in ns a
(ns a)
(s/defschema Kikka {:a s/Str})
; => #'a/Kikka

;; define a schema in ns b using schema from a
(ns b)
(require 'a)
(s/defschema Kukka {:kikka a/Kikka})
; => #'b/Kukka

;; modify the original schema
(ns a)
(s/defschema Kikka {:a s/Any})
; => #'a/Kikka

;; the schema-names are still the same, but the linked value is out of sync
(ns b)
(= (s/schema-name a/Kikka) 
   (s/schema-name (:kikka Kukka)))
; => true

(= a/Kikka
   (:kikka Kukka))
; => false
```

### DIFF
--- a/test/cljx/schema/core_test.cljx
+++ b/test/cljx/schema/core_test.cljx
@@ -1232,25 +1232,17 @@
 
 (deftest test-defschema
   (is (= 'TestFoo (:name (meta TestFoo))))
-  (is (= 'schema.core-test/TestFoo (:full-name (meta TestFoo)))))
+  (is (= 'schema.core-test (:ns (meta TestFoo)))))
 
 (deftest schema-with-name-test
   (let [schema (s/schema-with-name {:baz s/Num} 'Baz)]
     (valid! schema {:baz 123})
     (invalid! schema {:baz "abc"})
     (is (= 'Baz (s/schema-name schema)))
-    (is (=  nil (s/full-schema-name schema)))))
+    (is (=  nil (s/schema-ns schema)))))
 
 (deftest schema-name-test
   (is (= 'TestFoo (s/schema-name TestFoo))))
 
-(deftest full-schema-name-test
-  (is (= 'schema.core-test/TestFoo (s/full-schema-name TestFoo))))
-
-(s/defschema TestBar "BarTest" {:foo s/Str})
-
-#+clj ;; no Var meta-data on cljs
-(deftest defschema-var-metadata
-  (let [{:keys [line doc]} (-> TestBar s/full-schema-name resolve meta)]
-    (is (= line 1250))
-    (is (= doc "BarTest"))))
+(deftest schema-ns-test
+  (is (= 'schema.core-test (s/schema-ns TestFoo))))
--- a/test/cljx/schema/core_test.cljx
+++ b/test/cljx/schema/core_test.cljx
@@ -1231,13 +1231,26 @@
 (s/defschema TestFoo {:bar s/Str})
 
 (deftest test-defschema
-  (is (= 'TestFoo (:name (meta TestFoo)))))
+  (is (= 'TestFoo (:name (meta TestFoo))))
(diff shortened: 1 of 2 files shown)


ikitommi:
hmm - the tests do pass locally.

loganlinn:
@ikitommi We believe this is an issue with TravisCI. We've been seeing failures across several of our repos lately. [They are apparently working on the issue](https://twitter.com/traviscistatus/status/592902357144498176). I've restarted the tests to see if it's been resolved.

test/cljx/schema/core_test.cljx
```diff
@@ -1231,13 +1231,26 @@
...
+
+(deftest full-schema-name-test
+  (is (= 'schema.core-test/TestFoo (s/full-schema-name TestFoo))))
+
+(s/defschema TestBar "BarTest" {:foo s/Str})
+
+#+clj ;; no Var meta-data on cljs
+(deftest defschema-var-metadata
+  (let [{:keys [line doc]} (-> TestBar s/full-schema-name resolve meta)]
+    (is (= line 1250))
```
w01fe: :cool: but also fragile -- will have to be updated every time we change the tests.  Maybe just remove this?

w01fe:
Looks great overall.  Besides the one comment on the test, I'm not sure `full-name` is idiomatic -- it seems like either attaching `:ns`, or if you really want to store a full name I believe `qualified-name` might be the standard terminology -- thoughts?

w01fe:
(Also, thanks for the PR!  And I've been having trouble getting Travis to build properly on schema and plumbing, can't figure out why.  I just turned off the integration until I have time to look into it further)

ikitommi:
I think just having `:ns` is good, one can construct the qualified-name with that and the `:name` on the client projects. Will change to that and remove the last test.

w01fe:
Awesome, thanks!  Just bump the thread when it's updated and I'll get this merged.  Also let me know if you'd like me to cut a release once it's merged.

ikitommi:
Added a `schema-ns` like `schema-name` to hide from the internal implementation. Is this ok? Resolving the Var meta-data should work like this:

```clojure
(defschema Abba {:a Str})

(meta (ns-resolve (schema-ns Abba) (schema-name Abba)))
``` 
and release would be nice, also looking forward to the CLJS + Recursive :)

w01fe:
One more minor comment, otherwise looks good -- thanks again!

w01fe:
Awesome, thanks!

w01fe:
https://clojars.org/prismatic/schema/versions/0.4.2

ikitommi:
thanks!

PR #212, opened 2015-05-01
extends Schema for IPersistentMap instead of APersistentMap

ikitommi:
This PR gives wider support for custom Map-implementations. Not all Maps are extending the abstract `APersistentMap`, one is these is the `flatland.ordered.map.OrderedMap`.

with current version:

```clojure
lein try prismatic/schema "0.4.2" org.flatland/ordered

(require '[flatland.ordered.map :as m])
(satisfies? s/Schema (m/ordered-map {}))
; => false
``` 

with this modification:
```clojure
lein try prismatic/schema "0.4.3-SNAPSHOT" org.flatland/ordered

(require '[flatland.ordered.map :as m])
(satisfies? s/Schema (m/ordered-map {}))
; => true
```

related to https://github.com/metosin/compojure-api/issues/95

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -744,7 +744,7 @@
   (into {} (for [[k v] this] (vec (next (explain (map-entry k v)))))))
 
 (extend-protocol Schema
-  #+clj clojure.lang.APersistentMap
+  #+clj clojure.lang.IPersistentMap
   #+cljs cljs.core.PersistentArrayMap
   (walker [this] (map-walker this))
   (explain [this] (map-explain this))


Deraen:
This might be problematic as records implement IPersistentMap.

ikitommi:
Another way would be to open up `map-walker` and `map-explain`. Currently private.

w01fe:
I believe this was an intentional choice, so as to not include Records -- but I don't really remember the full line of reasoning.  I can try to work through it again, but if it's OK with you opening up `map-walker` and `map-explain` instead let's just do that.

ikitommi:
hi. both are good options, your lib, your call :)

but without the PR:

```clojure
lein try prismatic/schema "0.4.2"

(defrecord Kikka [])
(satisfies? s/Schema Kikka)
; => true
(s/check Kikka (Kikka.))
;  => nil
```

w01fe:
You're using Kikka as a class schema here; it's not behaving as a map
schema.

```clojure
(defrecord Foo [])
(s/explain Foo)
;; ==> user.Foo

(class (s/explain Foo))
;; ==> clojure.lang.Symbol
```

w01fe:
Closing for now -- we have some changes coming that should take care of this in a more general way.

ikitommi:
ok, looking forward to it.

PR #215, opened 2015-05-13
Fix broken link in README

immoh:
Link to details about controlling validation code generation points to wrong line number


w01fe:
Thanks!

PR #218, opened 2015-05-22
single-artifact fix for AOT bug

martinraison:
Here it is. This should fix Prismatic/plumbing#74. I ran the tests against Clojure 1.6.0 and 1.7.0-RC1, everything passes in both cases.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -74,10 +74,8 @@
 
    See the docstrings of defrecord, fn, and defn for more details about how
    to use these macros."
-  ;; don't exclude fn because of bug in extend-protocol, and def because it's not a var.
-  ;; for clj, only refer to what we need to initialize the namespace
-  #+clj (:refer-clojure :only [= -> second vals ffirst refer-clojure concat when])
-  #+cljs (:refer-clojure :exclude [Keyword Symbol defrecord defn letfn defmethod fn])
+  ;; don't exclude def because it's not a var.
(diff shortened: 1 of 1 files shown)


w01fe:
image

Thank you!! 

My one concern is that `emit-impl` is private, and as such could cease to exist or change arbitrarily at any point in the future, which would create a new compatibility issue.  

Two potential solutions that ensure forward compatibility: 
 - Check the version number instead: https://github.com/Prismatic/plumbing/blob/master/src/plumbing/core.cljx#L42
 - Rather than checking for the issue being fixed, check if it's broken.  Wrap the check in a try/catch, and assume it's fixed if an exception is thrown.  

Thoughts?  

Thanks again, much appreciated!

martinraison:
Thanks for your feedback!

I updated the request with a more robust check. I basically try to call extend-protocol with `fn` excluded, and revert the exclusion if an exception is thrown. Is that close to what you had in mind?

I usually prefer avoiding direct version checks, especially since in this case we would need to check the version qualifier as well (1.7.0-RC1 is fixed, 1.7.0-alpha1 is not). This may not play well with forks of the Clojure compiler that use different version names.

w01fe:
Agree a functionality check is better than a version check, and this looks like a safe way to do it.  Thanks again for your contribution, much appreciated! 

Did you also test with AOT to confirm that the root issue is fixed?    Either way, I'll merge now, cut a snapshot and ask for feedback on the plumbing thread, and then cut a real release as soon as we have confirmation that everything looks good.

PR #233, opened 2015-06-28
Remove inappropriate commas

estsauver:
There're a few inappropriate commas in the readme.


w01fe:
Thanks for the PR, but I don't find the commas inappropriate.

PR #237, opened 2015-07-21
Configurable maximum length for values in error messages

conan:
Sometimes it's useful to be able to see the value that failed validation in the output, but the hard-coded maximum length of 19 characters (< 20) can prevent this.  This PR makes that value configurable using the `set-max-value-length!` function in core, by changing the value of an atom in utils.  The atom defaults to 19, so this is not a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ Thus, each time you invoke `stamped-names`, Schema will perform validation.
 
 To reduce generated code size, you can use the `*assert*` flag and `set-compile-fn-validation!` functions to control when validation code is generated ([details](https://github.com/Prismatic/schema/blob/master/src/clj/schema/macros.clj#L181)).
 
+Schema will attempt to reduce the verbosity of its output by restricting the size of values that fail validation to 19 characters.  If a value exceeds this, it will be replaced by the name of its class.  You can adjust this size limitation by calling `set-max-value-length!`.
+
 Finally, we use validation with coercion for API inputs and outputs.  See the coercion section below for details.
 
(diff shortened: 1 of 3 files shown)


conan:
Tests run in both clj and cljs, no new tests included as this is not a bugfix and other utility code doesn't seem to have tests around it (plus it's so simple there's not really a need).

w01fe:
Thanks for the PR!  I think for this case a dynamic variable that you just call `set!` on might be more idiomatic (matching e.g. *warn-on-reflection*, *print-length*, etc).  Thoughts?  Otherwise looks good, thanks again!

conan:
Yeah, I'll admit to a bit of a gap in my understanding with things like `*warn-on-reflection*` - they seem to be created in Java by Clojure, and all Clojure code we run is actually nested inside a `binding` form that defines them, so we can call `set!` on them.  I don't know how to do this effectively with dynamic vars I've defined myself.  The `use-fn-validation` toggle in Schema is made with a protocol and type, which adds a lot of complexity to it, but that implementation appears to be chosen to improve the runtime performance of the toggle, which may be switched a lot of times very fast in code.  Changing the length of the error output isn't something that needs anywhere near that kind of performance (would it ever be set more than once in an application?), and so I chose the simpler option.

I'd be happy to adopt a more idiomatic approach, I'm not particularly tied to this one, but it works and I don't know any better!

w01fe:
Ah, I forgot about the being nested in a binding part -- that does seem like more trouble than it's worth.   Thanks!

PR #240, opened 2015-07-27
queue schemas

mmcgill:
This commit adds schema support for `PersistentQueue`. The following creates a schema that matches a persistent queue whose elements (if any) match the sub-schema `Int`:

```clojure
   (require [core.schema :as s])
   (def int-queue (s/queue s/Int))
```


w01fe:
:+1: thanks for your contribution!

PR #242, opened 2015-08-13
Making string-coercion-matcher more consistent with json-coercion-matcher

oliyh:
I see that `json-coercion-matcher` adds `keyword-enum-matcher` and `set-matcher` to `+json-coercions+`.

Also `+string-coercions+` includes `+json-coercions+`, while `string-coercion-matcher` adds `keyword-enum-matcher` to `+string-coercions+` but omits `set-matcher`.

This change brings `string-coercion-matcher` more into line with `json-coercion-matcher`.

Thanks

### DIFF
--- a/src/cljx/schema/coerce.cljx
+++ b/src/cljx/schema/coerce.cljx
@@ -130,4 +130,5 @@
      and long and doubles (JVM only) from strings."
   [schema]
   (or (+string-coercions+ schema)
-      (keyword-enum-matcher schema)))
+      (keyword-enum-matcher schema)
+      (set-matcher schema)))
--- a/test/cljx/schema/coerce_test.cljx
+++ b/test/cljx/schema/coerce_test.cljx
@@ -62,8 +62,8 @@
 
 (deftest string-coercer-test
   (let [coercer (coerce/coercer Generic coerce/string-coercion-matcher)]
-    (is (= {:b true :i 1 :n 3.0 :s "asdf" :k1 {1 :hi} :k2 :bye :e :a :u #uuid "550e8400-e29b-41d4-a716-446655440000"}
-           (coercer {:b "true" :i "1" :n "3.0" :s "asdf" :k1 {"1" "hi"} :k2 "bye" :e "a" :u "550e8400-e29b-41d4-a716-446655440000"})))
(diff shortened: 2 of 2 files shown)


w01fe:
Thanks for the PR!  `string-coercion-matcher` was really intended for use coercing, e.g., a map of query params or form params, so it doesn't include the affordances for structured data.  I suppose there's no harm in adding this, but would you mind sharing some details of your use case so we can make sure to have a solid mental model of what people are using it for?  Thanks!

oliyh:
Hello,

The use case is with https://github.com/frankiesardo/pedestal-swagger/ where the entire request is being coerced according to various schemata, but using the same coercion matchers (so some may be headers, form data, others may be json body).

Currently it uses `+string-coercions+` but I was intending to change that to `string-coercion-matcher` if this PR was accepted. I take your point about string coercions not really caring about structured data, but I thought as it already cared about other primitive types as per `+json-coercions+` and `keyword-enum-matcher` it was better to bring it into line.

I have realised that it would be nicer if coercion matchers could be more readily composed - I've got some ideas around this for which I may open a PR later - which would reduce the need for changes like this.

w01fe:
I see -- thanks for the context.  

The recommended thing in this context might be to use different coercers for each field, as we do in fnhouse:

https://github.com/Prismatic/fnhouse/blob/master/src/fnhouse/middleware.clj#L31

That said, this seems ok for now.  Look forward to seeing what you come up with around composition.  

Thanks again!

cichli:
> string-coercion-matcher was really intended for use coercing, e.g., a map of query params or form params, so it doesn't include the affordances for structured data.

I've needed to combine `string-coercion-matcher` and `json-coercion-matcher` when coercing the query params from a query string like `?foo=1&foo=2&foo=3` to a schema like `{:foo #{sc/Int}}`.

In that case, I found that `some-fn` worked well for composing them:

```clojure
user> (require '[schema.core :as sc])
nil
user> (require '[schema.coerce :as sco])
nil
user> (let [schema {:foo #{sc/Int}}
            coercer (sco/coercer schema (some-fn sco/string-coercion-matcher
                                                 sco/json-coercion-matcher))
            query-params {:foo ["1" "2" "3"]}]
        (coercer query-params))
{:foo #{1 3 2}}
```

PR #244, opened 2015-08-20
Declarative schema specs

w01fe:
Proposal for internal implementation changes for Schema 1.0.  

Exposes internals of schemas in a more declarative way, simplifying implementation of new schema types and applications.  

Ultra-bare-bones experimental implementations of generation is included, but will be fleshed out before a release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
  * **BREAKING** Changes to the core Schema protocol will break existing third-party schema tooling and schema types.
  * **BREAKING** Records coerced to an ordinary (non-record) map schema are now converted to maps, rather than retaining their record type.
  * **BREAKING** `s/either` no longer works with coercion.  We recommend switching to `s/conditional` or other options.
- * **Deprecate** `s/both` in favor of new `s/constrained`.
+ * **Deprecate** `s/both` in favor of improved `s/conditional`.
  * `s/pred` can more intelligently guess the predicate name
  * `record` schemas can now coerce values to corresponding record types.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 1.0.0-alpha1
- * New schema backend, which is faster, simpler, and more declarative, enabling more applications and simplifying tooling.  Users of built-in schema types should experience very little or no breakage, but tooling or custom schema types will need to be updated.
+ * New schema backend, which is faster, simpler, and more declarative, enabling more applications and simplifying tooling.  Users of built-in schema types should experience very little or no breakage, but tooling or custom schema types will need to be updated.   As a concrete example of an application that's enabled, schema now experimentally supports test-check style generation from schemas.
  * **BREAKING** Changes to the core Schema protocol will break existing third-party schema tooling and schema types.
  * **BREAKING** Records coerced to an ordinary (non-record) map schema are now converted to maps, rather than retaining their record type.
  * **BREAKING** `s/either` no longer works with coercion.  We recommend switching to `s/conditional` or other options.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.0.0-alpha1
+ * New schema backend, which is faster, simpler, and more declarative, enabling more applications and simplifying tooling.  Users of built-in schema types should experience very little or no breakage, but tooling or custom schema types will need to be updated.
+ * **BREAKING** Changes to the core Schema protocol will break existing third-party schema tooling and schema types.
+ * **BREAKING** Records coerced to an ordinary (non-record) map schema are now converted to maps, rather than retaining their record type.
+ * **BREAKING** `s/either` no longer works with coercion.  We recommend switching to `s/conditional` or other options.
+ * **Deprecate** `s/both` in favor of new `s/constrained`.
+ * `s/pred` can more intelligently guess the predicate name
+ * `record` schemas can now coerce values to corresponding record types.
+
 ## 0.4.4
  * Fix ClojureScript warnings about `map->Record` constructors being redefined.
  * Add queue schemas
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -426,9 +426,9 @@
     (variant/variant-spec
      spec/+no-precondition+
      (for [s schemas]
-       {:guard (complement (checker s))
+       {:guard (complement (checker s)) ;; since the guard determines which option we check against
         :schema s})
-     #(list 'matches-some-either-clause? %)))
+     #(list 'some-matching-either-clause? %)))
   (explain [this] (cons 'either (map explain schemas))))
 
 (clojure.core/defn either
@@ -500,7 +500,7 @@
      spec/+no-precondition+
      (for [[p s] preds-and-schemas]
        {:guard p :schema s})
-     #(list 'matches-some-condition? %))) ;; TODO: improve error message
+     #(list 'some-matching-condition? %))) ;; TODO: improve error message
   (explain [this]
     (->> preds-and-schemas
          (mapcat (clojure.core/fn [[pred schema]] [pred (explain schema)]))
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -443,25 +443,6 @@
   (Either. schemas))
 
 
-;;; constrained (matches precondition and schema)
-
-(clojure.core/defrecord Constrained [schema pred fn-name]
-  Schema
-  (spec [this]
-    (variant/variant-spec
-     (spec/precondition this pred #(list fn-name %))
-     [{:schema schema}]))
-  (explain [this] (list 'constrained (explain schema) fn-name)))
-
-(clojure.core/defn constrained
-  "A value that must satisfy both schema and precondition.
-   Replacement for `both` with cleaner semantics."
-  ([schema pred]
-     (constrained schema pred (symbol (utils/fn-name pred))))
-  ([schema pred pred-name]
-     (Constrained. schema pred pred-name)))
-
-
 ;;; both (satisfies this schema and that one)
 
 (clojure.core/defrecord Both [schemas]
@@ -484,7 +465,8 @@
 (clojure.core/defn ^{:deprecated "1.0.0"} both
   "A value that must satisfy every schema in schemas.
 
-   DEPRECATED: prefer 'constrained' instead.
+   DEPRECATED: prefer 'conditional' with a single condition
+   instead.
 
    When used with coercion, coerces each schema in sequence."
   [& schemas]
@@ -493,18 +475,24 @@
 
 ;;; conditional (choice of schema, based on predicates on the value)
 
-(clojure.core/defrecord ConditionalSchema [preds-and-schemas]
+(clojure.core/defrecord ConditionalSchema [preds-and-schemas error-symbol]
   Schema
   (spec [this]
     (variant/variant-spec
      spec/+no-precondition+
      (for [[p s] preds-and-schemas]
        {:guard p :schema s})
-     #(list 'some-matching-condition? %))) ;; TODO: improve error message
+     #(list (or error-symbol
+                (if (= 1 (count preds-and-schemas))
+                  (symbol (utils/fn-name (ffirst preds-and-schemas))))
+                'some-matching-condition?)
+            %)))
   (explain [this]
-    (->> preds-and-schemas
-         (mapcat (clojure.core/fn [[pred schema]] [pred (explain schema)]))
-         (cons 'conditional))))
+    (cons 'conditional
+          (concat
+           (mapcat (clojure.core/fn [[pred schema]] [(symbol (utils/fn-name pred)) (explain schema)])
+                   preds-and-schemas)
+           (when error-symbol [error-symbol])))))
 
 (clojure.core/defn conditional
   "Define a conditional schema.  Takes args like cond,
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.0.0-alpha1
+ * New schema backend, which is faster, simpler, and more declarative, enabling more applications and simplifying tooling.  Users of built-in schema types should experience very little or no breakage, but tooling or custom schema types will need to be updated.
```
davegolland: :sheep: might want to mention that we now experimentally support test-check style generation from schema as a concrete example of an application that's enabled

src/cljx/schema/core.cljx
```diff
@@ -473,47 +422,71 @@
...
-                   (list 'some (list 'check '% (utils/value-name x)) 'schemas))
-                  (let [res ((first sub-walkers) x)]
-                    (if-not (utils/error? res)
-                      res
-                      (recur (next sub-walkers)))))))))
+  (spec [this]
+    (variant/variant-spec
+     spec/+no-precondition+
+     (for [s schemas]
+       {:guard (complement (checker s))
```
davegolland: why is this guard here, i thought having guard would be equivalent? is it necessary for other applications besides checking?

seems like it is at least worth adding a small comment.

src/cljx/schema/core.cljx
```diff
@@ -473,47 +422,71 @@
...
+   bad error messages.  Please consider using `conditional` and friends
+   instead; they are more efficient, provide better error messages,
+   and work with coercion."
   [& schemas]
   (Either. schemas))
 
 
+;;; constrained (matches precondition and schema)
+
+(clojure.core/defrecord Constrained [schema pred fn-name]
```
davegolland: :sheep: is this the same as `conditional` with only 1 pred-schema pair and different error messages? not sure why it would be preferred to conditional?

src/cljx/schema/core.cljx
```diff
@@ -522,17 +495,16 @@
...
-            (clojure.core/fn [x]
-              (if-let [[_ match] (first (filter (clojure.core/fn [[pred]] (pred x)) preds-and-walkers))]
-                (match x)
-                (macros/validation-error this x (list 'matches-some-condition? (utils/value-name x)))))))
+  (spec [this]
+    (variant/variant-spec
+     spec/+no-precondition+
+     (for [[p s] preds-and-schemas]
+       {:guard p :schema s})
+     #(list 'matches-some-condition? %))) ;; TODO: improve error message
```
davegolland: :sheep: how about `no-matching-condition?` to mirror the error message from clojure's `case`: `IllegalArgumentException No matching clause`

src/cljx/schema/core.cljx
```diff
@@ -563,22 +535,17 @@
 
 (clojure.core/defrecord Recursive [derefable]
   Schema
-  (walker [this]
-          (let [a (atom nil)]
-            (reset! a (start-walker
-                       (let [old subschema-walker]
-                         (clojure.core/fn [s] (if (= s this) #(@a %) (old s))))
-                       @derefable))))
+  (spec [this] (variant/variant-spec spec/+no-precondition+ [{:schema @derefable}]))
```
davegolland: :+1: cool

src/cljx/schema/core.cljx
```diff
@@ -861,41 +785,36 @@
...
+  (spec [this]
+    (collection/collection-spec
+     (spec/precondition
+      this
+      (clojure.core/fn [x] (or (nil? x) (sequential? x) #+clj (instance? java.util.List x)))
+      #(list 'sequential? %))
+     vec
+     (let [[singles multi] (parse-sequence-schema this)]
+       (concat
+        (for [^One s singles]
```
davegolland: nice! this is very clean

src/cljx/schema/spec/core.cljx
```diff
@@ -0,0 +1,87 @@
...
+   These two use-cases aren't priveledged, just the two that are considered core
+   to being a Spec."
+  (subschemas [this]
+    "List all subschemas")
+  (checker [this params]
+    "Create a function that takes [data], and either returns a walked version of data
+     (by default, usually just data), or a utils/ErrorContainer containing value that looks
+     like the 'bad' parts of data with ValidationErrors at the leaves describing the failures.
+
+     params are: subschema-walker, return-walked?, and cache."))
```
davegolland: should it be `subschema-checker`?

also, maybe clarify what the parameters mean? 

maybe something like:
"params is a map specifying:
subschema-checker - a function for checking subschemas
returned-walked? - a boolean specifying whether to return the original or walked version of the data
cache - a map structure from schema to positive result"

src/cljx/schema/spec/core.cljx
```diff
@@ -0,0 +1,87 @@
...
+   If the datum passes the predicate, returns nil.
+   Otherwise, returns a validation error with description (err-f datum-description),
+   where datum-description is a (short) printable standin for the datum."
+  [s p err-f]
+  (fn [x]
+    (when-let [reason (macros/try-catchall (when-not (p x) 'not) (catch e# 'throws?))]
+      (macros/validation-error s x (err-f (utils/value-name x)) reason))))
+
+(defmacro simple-precondition
+  "A simple precondition where f-sym names a predicate (e.g. (simple-precondition s map?)"
```
davegolland: :sheep: mismatched parens

src/cljx/schema/spec/core.cljx
```diff
@@ -0,0 +1,87 @@
...
+      (wrap-recursive-delay (delay #+clj (.get cache cache-key) #+cljs (@cache cache-key)))
+      w)
+    (do #+clj (.put cache cache-key ::in-progress) #+cljs (swap! cache assoc cache-key ::in-progress)
+        (let [res (result-fn)]
+          #+clj (.put cache cache-key res) #+cljs (swap! cache assoc cache-key res)
+          res))))
+
+(defn sub-checker
+  [{:keys [schema error-wrap]}
+   {:keys [subschema-checker ^java.util.Map cache] :as params}]
```
davegolland: this typehint might not work for clojurescript

src/cljx/schema/spec/core.cljx
```diff
@@ -0,0 +1,87 @@
...
+(defn run-checker
+  [f return-walked? s]
+  (f
+   s
+   {:subschema-checker f
+    :return-walked? return-walked?
+    :cache #+clj (java.util.IdentityHashMap.) #+cljs (atom {})}))
+
+(defn with-cache [^java.util.Map cache cache-key wrap-recursive-delay result-fn]
+  (if-let [w #+clj (.get cache cache-key) #+cljs (@cache cache-key)]
```
davegolland: i'm not sure i understand the caching policy.. seems to only be a function of the schema, not of the input value.. i think i'm missing something, but can't this result in incorrect behavior if the same subschema appears twice in a composite schema?

or maybe there are assumptions made about when caches are created, in which case i think they should be documented

src/cljx/schema/spec/variant.cljx
```diff
@@ -0,0 +1,57 @@
...
+   [schema.spec.core :as spec])
+  #+cljs (:require-macros [schema.macros :as macros]))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Variant Specs
+
+(defn- option-step [o params else]
+  (let [g (:guard o)
+        c (spec/sub-checker o params)
```
davegolland: if there is no cache specified in the params, then i think the `spec/with-cache` might throw a NPE. are these intended to be the same params?

src/cljx/schema/spec/core.cljx
```diff
@@ -0,0 +1,87 @@
...
+    (if (= ::in-progress w) ;; recursive
+      (wrap-recursive-delay (delay #+clj (.get cache cache-key) #+cljs (@cache cache-key)))
+      w)
+    (do #+clj (.put cache cache-key ::in-progress) #+cljs (swap! cache assoc cache-key ::in-progress)
+        (let [res (result-fn)]
+          #+clj (.put cache cache-key res) #+cljs (swap! cache assoc cache-key res)
+          res))))
+
+(defn sub-checker
```
davegolland: when should the user call `sub-checker` and when should `run-checker` be used? maybe add some docstrings?

src/cljx/schema/spec/variant.cljx
```diff
@@ -0,0 +1,57 @@
...
+   The semantics of `options` is that the options are processed in order;
+   the datum must match the schema for the first option for which `guard`
+   passes."
+  ([pre options]
+     (variant-spec pre options nil))
+  ([pre ;- spec/Precondition
+    options ;- [{:schema (s/protocol Schema)
+    ;;           (s/optional-key :guard) (s/pred fn?)
+    ;;           (s/optional-key :error-wrap) (s/pred fn?)}]
+    err-f ;- (s/pred fn?)
```
davegolland: :sheep: what is err-f?

src/cljx/schema/utils.cljx
```diff
@@ -40,18 +45,29 @@
...
+  #+cljs (unmunge
+          (or (not-empty (second (re-find #"function ([^\(]*)\(" (str f))))
+              "function"))
+  #+clj (let [s (.getName (class f))
+              slash (.lastIndexOf s "$")
+              raw (unmunge
+                   (if (>= slash 0)
+                     (str (subs s 0 slash) "/" (subs s (inc slash)))
+                     s))]
+          (if (.startsWith raw "clojure.core/") (subs raw 13) raw)))
```
davegolland: :sheep: :sheep: :sheep: `(str/replace raw #"^clojure.core/" "")`

src/cljx/schema/utils.cljx
```diff
@@ -117,24 +133,6 @@
   (when (error? x)
     (.-error ^ErrorContainer x)))
 
-(defn wrap-error-name
```
davegolland: :hocho: :+1:

test/cljx/schema/core_test.cljx
```diff
@@ -229,6 +239,21 @@
...
+(deftest constrained-test
+  (let [schema (s/constrained
+                {s/Keyword s/Keyword}
+                (fn [m] (every? (fn [[k v]] (= k v)) m))
+                'equal-keys?)]
+    (valid! schema {})
+    (valid! schema {:foo :foo :bar :bar})
+    (invalid! schema {"foo" "foo"})
+    (invalid! schema {:foo :bar} "(not (equal-keys? {:foo :bar}))")
+    (invalid! schema {:foo 1} "(not (equal-keys? {:foo 1}))")
```
davegolland: oh, i see, the error messages are better

davegolland:
LGMC!

src/cljx/schema/spec/core.cljx
w01fe: This is just a cache of the checking function.  It serves two purposes: 
 - (1) speed up the creation of the checker when there are many nested occurrences of the same schema (someone ran into the exponential behavior on an earlier release)
 - (2) facilitate the recursive schema.

I'll add a comment.

test/cljx/schema/core_test.cljx
w01fe: yeah, but you're right that they could be unified.  I'm not sure if there's an easy way to get predicate names into conditional schema, that would also help with the error messages there though so is worth thinking about.  ideas?

src/cljx/schema/core.cljx
w01fe: Because the semantics of `variant` when checking is that you use the guard to select the schema, then the value **must** match that schema -- no backtracking.  A 'better' alternative would be to create another custom spec type for `either`, but i wasn't really inclined to make `either` more attractive :)

src/cljx/schema/core.cljx
w01fe: it will be wrapped in a `not` -- I'll change it to `some-matching-condition?`

src/cljx/schema/spec/core.cljx
w01fe: I wasn't sure exactly what to call it -- I wanted to make clear that you could pass something that wasn't just `checker` in here, which is how coercion works.  Suggestions?

w01fe: nevermind, i see it's just plain inconsistent.

src/cljx/schema/spec/variant.cljx
w01fe: I'm not sure I follow, the same params as what?  `cache` should always be in params.

w01fe:
Thanks for the comments!  Think I addressed them all in 0d673b5 modulo things I had questions about.

src/cljx/schema/spec/variant.cljx
davegolland: oh! nvm, i think i was getting confused between "options" and "params". i get it now, params are just for checking

test/cljx/schema/core_test.cljx
davegolland: i'm not totally sure. possibly something like spec/precondition instead of the predicates in the conditional? i guess that would be a breaking change.

not sure if it's too dirty, but if there is an odd number of args to the conditional, the last one is the error message?

davegolland:
:shipit:

src/cljx/schema/spec/collection.cljx
```diff
@@ -0,0 +1,85 @@
...
+   schematized.  At the top level, the collection has a precondition
+   (presumably on the overall type), a constructor for the collection from a
+   sequence of items, an element spec, and a function that constructs a
+   descriptive error on failure.
+
+   The element spec is a sequence of maps, each of which provides an element
+   schema, cardinality, parser (allowing for efficient processing of
+   structured collections), and optional error wrapper."
+  [pre ;- spec/Precondition
+   constructor ;- (s/=> s/Any [(s/named s/Any 'checked-value)])
```
davegolland: oh! :sheep: :sheep: why not derive the constructor from the schema? i.e. `#(into (empty s) %)`?

i know that this is less flexible, but in all real-world cases (vector, map, set) this works as the constructor. i guess it would be cleaner to have it be taken from the data value being checked instead of the schema..

also, it would support preserving the type of sequence (list, vector) and the type of map (i.e. sorted vs hashmap)?

w01fe: It's true mostly, but it doesn't always work -- in schema, for Records or
(now) queues, and externally I know people are using for other collections
where the schema may not match the data type.

The `empty` trick also mostly works on the datum, but not for some types
(such as records).

PR #248, opened 2015-08-21
introduces cond-pre to replace either

davegolland:
s/cond-pre behaves like s/either, taking a list of schemas to match. But it is a little cleaner because it creates an s/conditional where the predicates are based on the schema spec preconditions.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -431,9 +431,11 @@
      #(list 'some-matching-either-clause? %)))
   (explain [this] (cons 'either (map explain schemas))))
 
-(clojure.core/defn either
+(clojure.core/defn ^{:deprecated "1.0.0"} either
   "A value that must satisfy at least one schema in schemas.
    Note that `either` does not work properly with coercion
+
+   DEPRECATED: prefer `cond-pre`
 
    WARNING: either does not work with coercion.  It is also slow and gives
    bad error messages.  Please consider using `conditional` and friends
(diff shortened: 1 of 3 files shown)


src/cljx/schema/core.cljx
```diff
@@ -515,6 +487,83 @@
...
+(defprotocol HasPrecondition
+  (precondition [this]
+    "Return the Precondition for this schema.
+     A Precondition is a function of a value that returns a
+     ValidationError or nil (see spec.core)"))
+
+(extend-protocol HasPrecondition
+  schema.spec.leaf.LeafSpec
+  (precondition [this]
+    (:pre this))
```
w01fe: `(.-pre ^schema.spec.leaf.LeafSpec this)` is safer.  Same below.

src/cljx/schema/core.cljx
```diff
@@ -515,6 +487,83 @@
...
+
+  schema.spec.variant.VariantSpec
+  (precondition [this]
+    (let [{:keys [pre options]} this]
+      (some-fn
+       pre
+       (spec/precondition
+        Any
```
w01fe: i think this should be `this`

src/cljx/schema/core.cljx
```diff
@@ -515,6 +487,83 @@
...
+        (complement
+         (apply every-pred
+                (for [{:keys [guard schema]} options]
+                  (if guard
+                    (some-fn (spec/precondition schema guard #(list 'guard %))
```
w01fe: `schema` should probably also be `this`

src/cljx/schema/core.cljx
```diff
@@ -515,6 +487,83 @@
...
+(clojure.core/defn cond-pre
+  "A replacement for `either` that constructs a conditional schema
+   based on the schema spec preconditions of the component schemas.
+
+   EXPERIMENTAL"
+  [& schemas]
+  (apply conditional
+         (for [s schemas
+               t [(comp not (precondition (spec s))) s]]
+           t)))
```
w01fe: :sheep: might be clearer to do `(apply concat (for [s schemas] [(comp ...) s]))`

src/cljx/schema/core.cljx
```diff
@@ -515,6 +487,83 @@
...
+     A Precondition is a function of a value that returns a
+     ValidationError or nil (see spec.core)"))
+
+(extend-protocol HasPrecondition
+  schema.spec.leaf.LeafSpec
+  (precondition [this]
+    (:pre this))
+
+  schema.spec.variant.VariantSpec
+  (precondition [this]
```
w01fe: :goat: it seems like we're jumping through some hoops to make this a proper Precondition, which it isn't really.  Maybe it would be cleaner to just make it a simple predicate?

test/cljx/schema/core_test.cljx
```diff
@@ -277,6 +277,23 @@
       (is (= '(conditional odd? Int weird?)
              (s/explain (s/conditional odd? s/Int 'weird?)))))))
 
+(deftest cond-pre-test
+  (let [s (s/cond-pre
+           s/Int
+           (s/maybe s/Str)
+           (s/cond-pre s/Keyword {:x s/Int})
+           (s/both [s/Num] (s/pred (fn [xs] (even? (count xs))) 'even-len?)))]
```
w01fe: is this predicate really relevant to `cond-pre` in the test?  It might be moreso if there was another sequence schema afterwards.

w01fe:
:+1:

PR #253, opened 2015-08-21
Improve generators and completers

w01fe:


src/clj/schema/experimental/generators.clj
```diff
@@ -27,7 +27,7 @@
   [{:keys [schema]}
    {:keys [subschema-generator ^java.util.Map cache] :as params}]
   (spec/with-cache cache schema
-    (fn [d] (generators/make-gen (fn [r s] (generators/call-gen @d r s))))
+    (fn [d] (#'generators/make-gen (fn [r s] (generators/call-gen @d r (quot s 2)))))
```
davegolland: what does this do? why quote the symbol?

davegolland:
:+1:

src/clj/schema/experimental/generators.clj
w01fe: it was made private, this sidesteps that :(.  I wasn't sure how else to do this, I don't see any other public way to make a generator

PR #257, opened 2015-08-25
check the guard when generating from a variant schema

davegolland:


w01fe:
This was intentional, in that the `guard` of each option was meant to be a sub-property of the schema; so that it's basically just an optimization to allow greedy checking, but not part of the official semantics.

That said, I think I didn't document that, and the way we've set up `conditional` to work like `constrained` may not work with this (depending on if the constraint is put in the guard or the condition) -- I suppose this is safer, even if it's not strictly necessary if everything is setup right.  

Thoughts?  I'm fine with merging this, but also thing we should make the semantics of the guards clear either way.

davegolland:
Ah, didn't realize this was intentional. I wanted to add it because I was getting some invalid data generated from `(s/conditional pos? s/Int)` for the writeup.

w01fe:
Yeah, I think `conditional` could be rewritten to put the constraint into the precondition when there's onhly one so that it would work without changes, but it seems like an invitation for bugs so this is probably better.  What's your gut?  Either way, can you please add a sentence to the docstring of s/conditional clarifying the semantics of the guard?  Thanks!

davegolland:
@w01fe is this along the lines of what you mean: 1e11946

w01fe:
Looks good, thanks!

PR #265, opened 2015-09-03
Catch and report exceptions in guards like preconditions

### LINKED ISSUE #263: Matching of variant guards should catch exceptions
`(s/validate (s/conditional even? s/Int) "1")` 

throws an exception when it should behave the same as:

`(s/check (s/pred even?) "1")`

(returns an error like `(throws? (even? "1"))`)

w01fe:
Fixes #263

### DIFF
--- a/src/cljx/schema/spec/variant.cljx
+++ b/src/cljx/schema/spec/variant.cljx
@@ -13,7 +13,17 @@
   (let [g (:guard o)
         c (spec/sub-checker o params)
         step (if g
-               (fn [x] (if (g x) (c x) (else x)))
+               (fn [x]
+                 (macros/try-catchall
+                  (if (g x)
+                    (c x)
+                    (else x))
+                  (catch e#
+                      (macros/validation-error
+                       (:schema o)
+                       x
+                       (list (symbol (utils/fn-name g)) (utils/value-name x))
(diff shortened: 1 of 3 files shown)


src/cljx/schema/spec/variant.cljx
```diff
@@ -13,7 +13,17 @@
   (let [g (:guard o)
         c (spec/sub-checker o params)
         step (if g
-               (fn [x] (if (g x) (c x) (else x)))
+               (fn [x]
+                 (macros/try-catchall
+                  (if (g x)
```
loganlinn: is there merit to tightening the try-catch to be just around `(g x)` and not around sub-checker/else?

w01fe: Good call, will update.

w01fe: (It's just annoying since now we need trinary, unless you see another way around it?)

w01fe:
I guess an alternative is to just consider the guard false when it fails.  This will generally do the right thing, it just seems a bit gross and potentially more confusing in some cases.  Thoughts?

loganlinn:
Agree it's generally right behavior, but error messages could be
misleading. They way you have it now LGTM

PR #267, opened 2015-09-03
Move experimental files to cljx

### LINKED ISSUE #266: Completers are clojure only
Hey Prismatic, 

It seems like the new completers (which are awesome) are clojure only. It looks like they could be cross compiled to also be cljs compatible. Is there something keeping this Cljs only?

~Earl

estsauver:
Closes #266 .


w01fe:
Thanks.  But the files aren't renamed `.cljx`, so they probably aren't being compiled in either language.  

At minimum the JVM-specific stuff referring to classes and stuff will need to be made cross-platform, in addition to whatever test.check issues may or may not exist.

PR #270, opened 2015-09-15
Remove boolean(s) from clj primitives

### LINKED ISSUE #264: schema potentially tags boolean values as primitives
https://github.com/Prismatic/schema/blob/master/src/clj/schema/macros.clj#L72

Apparently this is incorrect? http://dev.clojure.org/jira/browse/CLJ-701?focusedCommentId=41088&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-41088

estsauver:
Resolves #264 I believe.

### DIFF
--- a/src/clj/schema/macros.clj
+++ b/src/clj/schema/macros.clj
@@ -69,8 +69,11 @@
     [(first s) (next s)]
     [nil s]))
 
-(def primitive-sym? '#{float double boolean byte char short int long
-                       floats doubles booleans bytes chars shorts ints longs objects})
+(if-cljs
+ (def primitive-sym? '#{float double boolean byte char short int long
+                        floats doubles booleans bytes chars shorts ints longs objects})
+ (def primitive-sym? '#{float double byte char short int long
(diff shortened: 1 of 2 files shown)


w01fe:
Thanks for your contribution!

While we're at it, I don't think the array types are valid in cljs, and i guess some of the other primitives (`int`, etc) probably aren't valid in either -- might be worth cleaning those up as well, if you're game.  Otherwise, we can just merge this and address that later.  Let me know what you think. 

Thanks again!

w01fe:
Actually, after further thought, I believe that the code is correct as-is for local bindings (e.g., `letk`; it's  functions arguments that restrict which primitive types can be used.)  I'm inclined to say this is just a documentation problem then; that we should be clear that you should only use primitive schemas on binding forms where Clojure would accept them as a valid type hint.  Thoughts?

w01fe:
#285 adds the discussed documentation updates.  Thanks again for submitting the PR; sorry for not thinking this through more before requesting this change.

PR #280, opened 2015-10-01
Add leaf generator for s/Uuid

benalbrecht:


src/clj/schema/experimental/generators.clj
```diff
@@ -116,6 +116,7 @@
     s/Symbol (generators/fmap (comp symbol name) generators/keyword)
     Object generators/any
     s/Any generators/any
+    s/Uuid (generators/fmap (fn [_] (java.util.UUID/randomUUID)) generators/int)
```
w01fe: This isn't correct -- you need to make sure you're using the random number generator internal to generators.  (They are supposed to be deterministic given the outer seed, and using a different RNG subverts that).  This thread shows some examples (https://gist.github.com/jcf/7a91c9473b57b1834704) although there's some question of whether they generate valid UUIDs -- I don't know too much about them.

w01fe:
Thanks for the PR!  As mentioned above, this isn't strictly correct.  If you want to propose a correct version that's generic (and hopefully, not too complicated), we'd be happy to add it.  Otherwise, if you don't really care about determinism, you should be able to use this version yourself without any changes to core (by passing it in as a leaf-generator).

benalbrecht:
Sure, i forgot about the seed. I changed the code to generate a type 3 UUID using `UUID/nameUUIDFromBytes`.
This function receives a byte array generated by `test.check.generators` as input.
Since this type of uuid is basically an md5 hash [1], supplying the same byte array results in the same uuid.

[1] https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_3_.28MD5_hash_.26_namespace.29

w01fe:
Looks good, thanks!

PR #281, opened 2015-10-02
Propagate return value of defn

### LINKED ISSUE #274: s/defn returns nil instead of var
```clojure
user> (s/defn f :- s/Int [x :- s/Int] (inc x))
nil
```
I would expect it to return `#'user/f` instead of nil. Now I cannot use `s/defn` in conjunction with `clojure.test/with-test`, because that macro expects the enclosed definition to return a new var to test.

loganlinn:
Match the return value of `schema.core/defn` with `clojure.core/defn`

Note: `let` the value rather than simply returning `(var ~name)` because some
versions of ClojureScript do not support vars.

Fixes #274


w01fe:
+:1:

Issue #269, opened 2015-09-11
Issue replacing `both` with `conditional`; a replacement for schema conjunction


lvh:
Imagine I want a schema for validating some URLs. In addition to checking if the URL is valid, I want to verify some other properties, for example, if it's scheme is in a supported set.

Previously, I did something like [this](https://github.com/lvh/icecap/blob/master/src/icecap/handlers/http.clj#L23-L24):

```clojure
(defn scheme
  "Get the scheme of a URL."
  [^String url]
  (.getScheme (URI. url)))

(defn valid-scheme?
  "Checks if the given URL has a valid scheme."
  [^String url]
  (#{"http" "https"} (scheme url)))

(def schema
    {:url (s/both sc/URI
                  (s/pred valid-scheme? 'valid-scheme?))})
```

Here, `sc` is [schema-contrib](https://github.com/sfx/schema-contrib); `sc/URI` is a schema for URIs.

`both` being deprecated, I tried to replace it with `conditional` as suggested. Fortunately `valid-scheme?` is already a predicate, so perhaps the obvious way to fix this is:

```clojure
{:url (s/conditional valid-scheme? sc/URI)
   :method (s/enum :GET :POST :DELETE :PUT :PATCH :HEAD)}
```

However, this raises some issues:

- The error message for this is worse. When adding a totally bogus URL (let's say, an integer), you would previously get `:url (throws? (URI 1))`; now you get `:url (throws? (icecap.handlers.http/valid-scheme? 1)`. While `1` indeed doesn't have a valid scheme, that's not quite what's wrong here.
- When it is actually just the scheme, the namespace that the pred is in is now added to the error message.
- `conditional` is a disjunction of conjunctions, whereas `both` (despite the name) allows *n* conjunctive schemas, not just two. How do you replace `both` for that case? `(conditional p1 (conditional p2 (conditional p3 schema))` seems gauche.

lvh:
For now, I ended up going with:

```clojure
(s/conditional valid-url?
                       (s/conditional valid-scheme?
                                      sc/URI))
```

Which is almost as good as the old `pred` solution except there doesn't seem to be an obvious way to specify a name for the predicate. Specifically, right now it uses the fully qualified name of the predicate, with namespace; the namespace is irrelevant to the consumer of the error.

w01fe:
Thanks for the context.  We're still thinking about adding an alternative to conditional that provides for a postcondition, rather than precondition, which could behave more like how `both` was sometimes used.  

[...]

PR #292, opened 2015-10-22
Add constrained schema with postconditions

### LINKED ISSUE #291: `both` -> `conditional` migration issue - different coercion behaviour
Strongly related to #269; feel free to close if this is effectively a duplicate.

We've hit an issue with coercion while migrating our codebase from `both` to `conditional`. Previously, we could do the following (where `schema.core` and `schema.coerce` are aliased to `sc` and `sco` respectively):

```clojure
user> (let [schema (sc/both sc/Int (sc/pred pos?))
            coercer (sco/coercer schema sco/string-coercion-matcher)]
        (coercer "1"))
1
```

I haven't figured out a way to accomplish the same thing with `conditional`:

```clojure
user> (let [schema (sc/conditional pos? sc/Int)
            coercer (sco/coercer schema sco/string-coercion-matcher)]
        (coercer "1"))
#schema.utils.ErrorContainer{:error (throws? (pos? "1"))}
```

Would the suggested post-condition variant of `conditional` suggested in #269 resolve this issue?

w01fe:
Solves the error message and coercion issues people were seeing after moving from `(s/both s (s/pred ...))` to `conditional`.  

Fixes #291 and #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
  * Add `coercer!` which throws on error
  * Add leaf generators for UUIDs
  * Make `s/defn` compatible with `with-test` 
+ * Add `constrained` schema for postconditions (replaces `(both x (s/pred ...))`)
 
 ## 1.0.1
  * Catch and report exceptions in guards the same as preconditions, rather than allowing them to propagate out.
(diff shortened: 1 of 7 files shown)


loganlinn:
LGTM :+1:

Issue #219, opened 2015-05-24
Switching to reader conditionals


danielcompton:
Is there any interest in switching to 1.7's Reader Conditionals once they have been out for a while? There is the very real issue of supporting previous versions of Clojure which wouldn't work with Reader Conditionals. What's the plan for supporting older versions of Clojure?

w01fe:
I think it would be prudent to wait until 1.7 has been out at least a year before starting this discussion, especially since AFAICT there's not much to gain from switching (besides simpler tooling for library maintainers) -- am I missing something there? 

 That said, if someone releases a tool that can produce a pre-1.7 compatible artifact from reader conditionals (essentially cljx updated for the new syntax) we could talk about switching sooner.

danielcompton:
Simpler tooling was my main idea, although it is a pretty minor benefit when cljx is already setup.

enaeher:
I'd like to make the case for switching to `cljc` sooner.

- cljx is deprecated by its author and no longer maintained. As a potential contributor I'd prefer not to spend time learning the idiosyncrasies of a tool which has already been deprecated.
- Jump-to-source IDE functionality like Cider's `cider-find-var` works correctly with reader conditionals but not with cljx. I'm also running into problems where sometimes the cljx nrepl middleware stops working and I can't compile files in Cider at all. (I'm sure this latter issue is fixable, but again, it feels like a waste of time to learn to fight a tool which has been superseded.)

w01fe:
Those are fine points, but you haven't said anything about how to address pre-1.7 support.   Do you have anything to add there?

danielcompton:
The only viable option I know of would be maintaining two release branches for a year and merging in patches from the 1.7 branch into 1.6. The cost benefit ratio doesn't seem worth it to me, but others may disagree.

enaeher:
I don't know your requirements. If backward compatibility is a hard requirement, then you're right, there is no choice but to stay with cljx. (But given the prominent bold-faced warning in the README that this is alpha software and that APIs will change, I don't think breaking backward compatibility would catch anyone by surprise. People who didn't want to upgrade Clojure could continue using their current version of Schema. Only people who want to upgrade Schema but not Clojure would be affected.)

Even if it's premature to drop 1.6 support now, a year is a long time. The 2014 State of Clojure survey showed 90% of respondents using 1.6 roughly six months after its release (though some of those also had systems running 1.5).

[...]

PR #299, opened 2015-11-10
Update profile for Clojure 1.8 to RC1

loganlinn:


w01fe:
:+1:

atroche:
Awesome! When 1.8 is released, are you guys planning to [move to reader conditionals](https://github.com/Prismatic/schema/issues/219) and add [CLJS support for generators and completers](https://github.com/Prismatic/schema/issues/266)?

w01fe:
We'll do this breaking release at some point, but doing them now would only give ~6 months backwards compatibility for Clojure and ~3 months for ClojureScript, which is less than we'd like.  If you want to use completers in cljs now, you can check out the fork  by @estsauver https://github.com/estsauver/schema/blob/master/test/cljx/schema/core_test.cljx#L44

atroche:
Okay, thanks!

On 12 November 2015 at 03:45, Jason Wolfe <notifications@github.com> wrote:

> We'll do this breaking release at some point, but doing them now would
> only give ~6 months backwards compatibility for Clojure and ~3 months for
> ClojureScript, which is less than we'd like. If you want to use completers
> in cljs now, you can check out the fork by @estsauver
> <https://github.com/estsauver>
> https://github.com/estsauver/schema/blob/master/test/cljx/schema/core_test.cljx#L44
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Prismatic/schema/pull/299#issuecomment-155839915>.
>



-- 
-- Alistair

PR #302, opened 2015-11-20
Add `Dunno` and `Approx` utilities

MerelyAPseudonym:
(The justification for these additions is in [this commit](https://github.com/MerelyAPseudonym/schema/commit/0bea7becba88d328bf372c1d12f878bd85030389?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8))

Assuming you even like the idea of `Dunno` and/or `Approx`, I'm totally open to polishing the PR. Just let me know.


w01fe:
Thanks for the PR -- I think the idea is interesting, but we want to keep this core project as simple as possible, so they might be a better fit for a "contrib"-type library.

PR #303, opened 2015-11-21
Upgrade test.check, start using fancier generators

gfredericks:
The biggest change here is that the numeric generators will now generate
a much larger range of numbers, which could be bad for anybody who was
relying on the default -200→200 range.

The tests are currently failing because apparently the infinite doubles can't be trivially cast to infinite floats, but that could be fixed with a bit of fiddling, once somebody confirms that these changes are otherwise acceptable.

### DIFF
--- a/src/clj/schema/experimental/generators.clj
+++ b/src/clj/schema/experimental/generators.clj
@@ -86,24 +86,16 @@
   "A mapping from schemas to generating functions that should be used."
   (s/=> (s/maybe Generator) Schema))
 
-(def gen-rational
-  "Simple generator of rational numbers."
-  (generators/one-of
-   [generators/int
-    (generators/fmap
-     (fn [[a b]] (/ (* a a a a a) b))
-     (generators/tuple
-      generators/int
-      generators/s-pos-int))]))
-
 (def +primitive-generators+
-  {Double (generators/fmap double gen-rational)
-   Float (generators/fmap float gen-rational)
-   Long generators/int
-   Integer (generators/fmap unchecked-int generators/int)
-   Short (generators/fmap unchecked-short generators/int)
-   Character (generators/fmap unchecked-char generators/int)
-   Byte (generators/fmap unchecked-byte generators/int)
+  {Double generators/double
+   ;; using unchecked-float here will unfortunately generate a lot of
+   ;; infinities, since lots of doubles are out of the float range
+   Float (generators/fmap unchecked-float generators/double)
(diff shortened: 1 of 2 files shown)


src/clj/schema/experimental/generators.clj
```diff
@@ -88,22 +88,19 @@
...
-   [generators/int
-    (generators/fmap
-     (fn [[a b]] (/ (* a a a a a) b))
-     (generators/tuple
-      generators/int
-      generators/s-pos-int))]))
+  (generators/fmap (fn [[num den]] (/ num den))
+                   (generators/tuple generators/large-integer
+                                     (generators/such-that #(not (zero? %))
+                                                           generators/large-integer))))
```
w01fe: This can be deleted, I think.

w01fe:
Thanks -- this is great!  Changes look good to me.

src/clj/schema/experimental/generators.clj
gfredericks: oh are we not supporting a direct Ratio schema? I guess I've never seen one...

w01fe: Nope -- this was just there to stand in for `generators/double`

gfredericks:
Okay, I removed `gen-rational` and fixed the tests -- they were failing because we were using `(fmap float ...)` instead of `(fmap unchecked-float ...)`, and `float` was catching the out-of-range numbers that are now being generated. The downside (as indicated in a comment) is that a significant portion of `(fmap unchecked-float generators/double)` will be infinities. This could be improved through some sort of trickery, but I figured for now it's fine since we're still generating decent floats pretty often enough and who whe hell uses `Float` instead of `Double` anyhow.

gfredericks:
I should mention that I couldn't get the CLJS tests to run, so somebody who's already setup to run those should double-check that everything builds okay.

gfredericks:
I should also stress that these changes mean that test.check >=0.9.0 is required (at least to compile the generatorts file), which means that clojure >=1.7.0 is required. Not sure what this means for users on older clojure versions who don't care about the generators features.

w01fe:
Looks good to me -- all tests pass (including cljs), just not on 1.6.  Given that it's experimental and "bring-your-own-dep", I don't think the version requirement is a concern either -- people with earlier Clojure versions simply won't be able to use the generators.  I'll add an entry to that effect to the changelog.  Thanks again!

PR #305, opened 2015-11-22
Better performance for anonymous schematized functions, via lazy checker creation

w01fe:
Fixes https://github.com/Prismatic/plumbing/issues/92

Unfortunately I don't know of a way to hoist the schema checker creation for anonymous functions, so it was happening on every creation instance.  This PR adds a delay around checker creation (and makes another small optimization), making it about 20 times faster to create anonymous functions (or `fnks`) than previously.  Should also speed up load time for schematized codebases slightly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 1.0.4
  * Update generators, bump minimum version of test.check to 0.9.0.  Generators will only work under Clojure 1.7.0+.
+ * Better performance for anonymous schematized functions, via lazy checker creation
 
 ## 1.0.3
  * Fix warning about overriding `atom` under Clojure 1.7
--- a/src/clj/schema/macros.clj
+++ b/src/clj/schema/macros.clj
@@ -211,8 +211,8 @@
         compile-validation (compile-fn-validation? env fn-name)]
     {:schema-binding [input-schema-sym (input-schema-form regular-args rest-arg)]
      :more-bindings (when compile-validation
-                      [input-checker-sym `(schema.core/checker ~input-schema-sym)
-                       output-checker-sym `(schema.core/checker ~output-schema-sym)])
+                      [input-checker-sym `(delay (schema.core/checker ~input-schema-sym))
+                       output-checker-sym `(delay (schema.core/checker ~output-schema-sym))])
      :arglist bind
      :raw-arglist original-arglist
      :arity-form (if compile-validation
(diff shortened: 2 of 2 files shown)


bfabry:
so this shifts the time creating the checkers from fn-creation-time to fn-first-execution-time right? what's the situation where that's useful? asking because attaching mutable things like Delay's to a fn makes them hard to serialize, as in https://github.com/Prismatic/schema/issues/308

w01fe:
The main reason for the change is that if you use `s/fn` (or `plumbing.core/fnk`), we have to create the checker each time we create an anonymous function (since there's no way I know of to lift it outside the body of the containing form to make it happen just once).  This makes these expressions 20x slower than they need to be, even when validation is not being used.  If you can think of an alternative that has the same effect, we'd be happy to consider it -- sorry for the difficulty.

bfabry:
Now that I look at it more closely, I'm actually quite confused by the delay, it seems like it is created and derefed at the same time, and so shouldn't make any difference to performance at all.
```
(def a-sym `(do (println "thunk") 5))
=> #'user/a-sym
`[~a-sym]
=> [(do (clojure.core/println "thunk") 5)]
(def a-sym `(delay (do (println "thunk") 5)))
=> #'user/a-sym
`[@~a-sym]
=> [(clojure.core/deref (clojure.core/delay (do (clojure.core/println "thunk") 5)))]

```

w01fe:
That's not what's going on though -- the `delay` is bound outside the fn, and deref'd inside.

```clojure
user> (pprint (macroexpand '(s/fn [])))
(let*
 [ufv__
  schema.utils/use-fn-validation
  output-schema9819
  schema.core/Any
  input-schema9820
  []
  input-checker9821
  (clojure.core/delay (schema.core/checker input-schema9820))
  output-checker9822
  (clojure.core/delay (schema.core/checker output-schema9819))]
 (schema.core/schematize-fn
  (clojure.core/fn
   fn9818
   ([]
    (clojure.core/let
     [validate__8342__auto__ (.get_cell ufv__)]
     (clojure.core/when
      validate__8342__auto__
      (clojure.core/let
       [args__8343__auto__ []]
       (clojure.core/when-let
        [error__8344__auto__ (@input-checker9821 args__8343__auto__)]
        (schema.macros/error!
         (schema.utils/format*
          "Input to %s does not match schema: %s"
          'fn9818
          (clojure.core/pr-str error__8344__auto__))
         {:schema input-schema9820,
          :value args__8343__auto__,
          :error error__8344__auto__}))))
     (clojure.core/let
      [o__8345__auto__ (clojure.core/loop [])]
      (clojure.core/when
       validate__8342__auto__
       (clojure.core/when-let
        [error__8344__auto__ (@output-checker9822 o__8345__auto__)]
        (schema.macros/error!
         (schema.utils/format*
          "Output of %s does not match schema: %s"
          'fn9818
          (clojure.core/pr-str error__8344__auto__))
         {:schema output-schema9819,
          :value o__8345__auto__,
          :error error__8344__auto__})))
      o__8345__auto__))))
  (schema.core/->FnSchema output-schema9819 [input-schema9820])))
```

bfabry:
ah, I was reading the value of :more-bindings as a let binding form, when it's actually creating a vector used elsewhere. will have to dig deeper, cheers

PR #306, opened 2015-11-24
Update changelog and project.clj to reflect new generator support

w01fe:
Everything (except generators) should still work fine under 1.6, but I think it would require some fancy footwork to keep running the tests under 1.6, so I just removed it for now.  Thoughts?


loganlinn:
sgtm

PR #311, opened 2015-11-30
Move error wrap fn out of schema cache

### LINKED ISSUE #310: Schema caching issue when using error-wrap
When using the optional error-wrap when making my own schema with collection-spec I ran into a bug in sub-checker. Sub-checker caches a checker function for later reuse but it includes the error-wrap call in the cached valued but not in the cache key. This means that when you reuse the same schema but with different error-wrap functions you get weird results that depend on what gets cached first.

Here is some clojure code that illustrates the problem:
```
(ns cache-test
  (:require [schema.core :as s]
            [schema.spec.core :as spec]
            [schema.spec.collection :as collection]
            [schema.utils :as utils])
  (:import (schema.core Schema)))

(defrecord ItemTest [first second])

(defrecord CacheTest [schema]
  Schema
  (spec [this]
    (collection/collection-spec
      (let [p (spec/precondition this #(instance? ItemTest %) #(list 'instance? ItemTest %))]
        (if-let [evf (:extra-validator-fn this)]
          (some-fn p (spec/precondition this evf #(list 'passes-extra-validation? %)))
          p))
      (fn [x] x)
      [{:schema     s/Str
        :cardinality :exactly-one
        :parser     (fn [item-col m]
                      (item-col (:first m))
                      m)
        :error-wrap (fn [err] [:first (utils/error-val err)])
        }
       {:schema      schema
        :cardinality :exactly-one
        :parser      (fn [item-col m]
                       (item-col (:second m))
                       m)
        :error-wrap  (fn [err] [:second (utils/error-val err)])
        }
       {:schema s/Any
        :cardinality :exactly-one
        :parser (fn [_ _] nil)}]
      (fn [_ elts _] elts)))
  (explain [_]
    (list 'cache-test)))


; Expect result to be [(error-val [:first '(not (instance? java.lang.String 12))])
;                      (error-val [:second '(not (instance? java.lang.String nil))])]
; Actual result is [(error-val [:second '(not (instance? java.lang.String 12))])
;                   (error-val [:second '(not (instance? java.lang.String nil))])]
(s/check (->CacheTest s/Str) (->ItemTest 12 nil))

; Expect result to be [(error-val [:first '(not (instance? java.lang.String 12))])
;                      nil]
; Actual result is [(error-val '(not (instance? java.lang.String 12)))
;                   nil]
(s/check (->CacheTest [s/Str]) (->ItemTest 12 nil))

; Expect result to be [(error-val [:first '(not (instance? java.lang.String 12))])
;                      (error-val [:second ['(not (instance? java.lang.String nil))])]
; Actual result is [(error-val '(not (instance? java.lang.String 12)))
;                   (error-val [:second ['(not (instance? java.lang.String nil))])]
(s/check (->CacheTest [s/Str]) (->ItemTest 12 [nil]))
```

The problem is that a checker function for s/Str gets cached based on the second schema in the collection and then that checked function with whatever error-wrap was on it gets called when checking the first schema. For the first test the error-wrap from the second schema is used and for the other tests there is no error-wrap in the cached value because the schema [s/Str] doesn't define an error-wrap for s/Str

griff:
I have fixed #310 as you sugested with tests to prevent a future regression.

### DIFF
--- a/src/cljx/schema/spec/core.cljx
+++ b/src/cljx/schema/spec/core.cljx
@@ -87,14 +87,13 @@
    Handles caching and error wrapping behavior."
   [{:keys [schema error-wrap]}
    {:keys [subschema-checker cache] :as params}]
-  (with-cache cache schema
-    (fn [d] (fn [x] (@d x)))
-    (fn []
-      (let [sub (subschema-checker schema params)]
-        (if error-wrap
-          (fn [x]
-            (let [res (sub x)]
-              (if-let [e (utils/error-val res)]
-                (utils/error (error-wrap res))
-                res)))
-          sub)))))
+  (let [sub (with-cache cache schema
+              (fn [d] (fn [x] (@d x)))
+              (fn [] (subschema-checker schema params)))]
+    (if error-wrap
+      (fn [x]
+        (let [res (sub x)]
+          (if-let [e (utils/error-val res)]
+            (utils/error (error-wrap res))
+            res)))
+      sub)))
--- a/test/cljx/schema/core_test.cljx
+++ b/test/cljx/schema/core_test.cljx
@@ -8,13 +8,15 @@
     - (invalid-call! s x) asserts that calling the function throws an error."
   #+clj (:use clojure.test [schema.test-macros :only [valid! invalid! invalid-call!]])
   #+cljs (:use-macros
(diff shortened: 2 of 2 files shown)


w01fe:
Looks good, thanks for the report and fix!

PR #319, opened 2015-12-17
support vars

pleasetrythisathome:
Provides support for Vars. Defines schema with subschema as vars can provide substantial printing improvements as well as other potentially useful info without changing anything about the schema format.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -244,6 +244,14 @@
   "Any value, including nil."
   (AnythingSchema. nil))
 
+;; Vars can contain schema
+
+(extend-protocol Schema
+  #+clj clojure.lang.Var #+cljs cljs.core.Var
+  (spec [this]
+    (spec (deref this)))
+  (explain [this]
+    (explain (deref this))))
 
 ;;; eq (to a single allowed value)
 
--- a/test/cljx/schema/core_test.cljx
+++ b/test/cljx/schema/core_test.cljx
@@ -95,6 +95,10 @@
   (valid! s/Any :whatever)
   (is (= 'Any (s/explain s/Any))))
 
+(deftest var-test
+  (valid! #'s/Any 10)
+  (is (= 'Any (s/explain #'s/Any))))
+
 (deftest eq-test
   (let [schema (s/eq 10)]
     (valid! schema 10)


w01fe:
Thanks for the PR.  Can you please explain the motivation behind this change, ideally with a concrete example?

pleasetrythisathome:
I find myself frequently using pretty deeply nested schema for data structures, and printing them is an absolute mess. Defining them as vars causes them to print as the var reference, which is more informative about what the schema actually is without having to define a custom schema type with a specific explain method.

I've also been thinking about generating documentation for functions from schema, and using vars would let me source map the schema and nested schemas to their original location!

```clojure
(s/defn phone-number? :- (s/maybe s/Str)
  [s :- (s/maybe s/Str)]
  (when s
    (re-matches #"\+1\d{10,10}" s)))

(s/defschema PhoneNumber
  (s/pred phone-number?))

(s/defschema Contact
  {:contact/uid s/Str
   :phone/number PhoneNumber})

(s/check Contact {:contact/uid "test"
                  :phone/number "222.333.4444"})
;; =>
;; {:phone/number (not (populace.utils.schema$eval91195/phone-number?--91200 "222.333.4444"))}

(pprint Contact)
;; =>
;; {:contact/uid java.lang.String,
;;  :phone/number
;;  {:p? #function[populace.utils.schema/eval91195$phone-number?--91200],
;;   :pred-name populace.utils.schema$eval91195/phone-number?--91200}}

(s/defschema Contact
  {:contact/uid s/Str
   :phone/number #'PhoneNumber})

(s/check Contact {:contact/uid "test"
                  :phone/number "222.333.4444"})
;; validates identitically
;; =>
;; {:phone/number (not (populace.utils.schema$eval91195/phone-number?--91200 "222.333.4444"))}

(pprint Contact)
;; printing shows the var instead of the schema
;; =>
;; {:contact/uid java.lang.String,
;;  :phone/number #'populace.utils.schema/PhoneNumber}

(pprint (meta (:phone/number Contact)))
;; preserves meta on original schema. possible to generate documentation with source mapping, etc.
;; =>
;; {:line 65,
;;  :column 0,
;;  :file
;;  "/Users/pleasetrythisathome/Code/work/populace/populace/src/populace/utils/schema.cljc",
;;  :doc "",
;;  :name PhoneNumber,
;;  :ns #namespace[populace.utils.schema]}
```

ikitommi:
for the documentation part, you can already resolve schema var from a meta of defschema'd schema, see https://github.com/metosin/schema-tools/blob/master/src/schema_tools/core.cljc#L237-L243

pleasetrythisathome:
@ikitommi that's cool. will definitely use that at some point as well

w01fe:
Thanks for the explanation and example, that's really helpful.  

I understand the motivation now, but my initial feeling is that this doesn't really hang together with the rest of the design of Schema -- I would expect a Var containing a schema `s` to match a Var containing something that matches `s`.   Thoughts?

pleasetrythisathome:
Not sure I follow, can you give a short example of what you'd expect?

w01fe:
What I mean is, every other place where we extend the Schema protocol to an existing type, the values that match such Schemas are in some way related to that type (maps, sets, Classes, etc).  So, for example, I might expect something like:

```clojure
(defn foo String)
(defn bar "1")
(s/validate #'foo #'bar)
```  

I'm not sure that's something we'd actually want to add in the future, but I'd be hesitant to rule it out by reserving `Var` for the use case you show above.  Does that make sense?

w01fe:
Closing for inactivity, please feel free to reopen if you want to discuss further.  Thanks!

PR #335, opened 2016-02-05
Add note about set-fn-validation! in README.md

hyPiRion:
Hi!

I find it valuable to experiment with validation during development to verify assumptions of different functions and their function types. However, how one enable/disable validation during development is not described in the readme. As a result, I spent an embarrasingly long time looking for it, as I did not assume it was easily available.

I think it may be worthy to put a small note about `set-fn-validation!` in the readme to make it easier for others to find it as well.


w01fe:
Thanks for the PR!

PR #342, opened 2016-02-22
Fix sequence generation (fixes #321)

### LINKED ISSUE #321: Sequence generators with optional elements don't follow sequence schema semantics
If a sequence generator declines to generate a value from an optional One schema, it still goes on to try generating values from the remaining subschemas inside the sequence schema, triggering errors from the validator embedded in the generator.

You can verify this by generating values from `FancySeq` as defined in the readme.

w01fe:
The existing `:cardinality` key was not expressive enough to capture the semantics of sequence schemas (namely, that if you don't generate an `s/optional` element, you have to stop there.  

This PR replaces `:cardinality` with a nested structure of optional tails that can properly express the semantics of existing collection schemas for the purpose of generation.  It also slightly generalizes the types of schemas that can be expressed with the collection spec, although probably not in a way that's useful to anyone.

The changes feel a bit heavyweight for what's accomplished, but I tried a few alternatives and couldn't get them to work without reasonable performance (and really didn't feel good about the option of adding another pseudo-`:cardinality` that expressed the missing constraint.  Suggestions for alternative approaches very welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.0
+ * **BREAKING** change the internal details of collection specs (should only be an issue for custom collection schemas that don't rely on helpers `one-element` or `all-elements`).
+ * Fix generation for sequence schemas containing a non-trailing `s/optional` element.
+
 ## 1.0.6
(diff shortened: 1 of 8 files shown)


w01fe:
Anyone want to take a look / comment?  I'd like to get this merged so we can finish splitting out generators as well.  Thanks!

PR #368, opened 2016-08-31
Add map entry constructor that returns values of the correct type

imrekoszo:
The other day I was tracking down a weird bug in a clojure project and narrowed it down to the following inconsistency:

```clojure
(type (first {:asdf "asdf"}))
=> clojure.lang.MapEntry

(type ((schema.coerce/coercer
         (schema.core/map-entry (schema.core/eq :asdf) sc/Str)
         (constantly nil))
        (first {:asdf "asdf"})))
=> clojure.lang.PersistentVector
```

Given a coercer with a coercion matcher that always returns nil, map entries are being changed into vectors. I would assume such a nil coercer to always return the unaltered input or something very close to that.

Added a test to force the code change and updated the contstructor the MapEntry schema uses.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -747,13 +747,17 @@
   (or (required-key? ks)
       (optional-key? ks)))
 
+(clojure.core/defn map-entry-ctor [[k v :as coll]]
+  #+clj (clojure.lang.MapEntry. k v)
+  #+cljs (vec coll))
+
 ;; A schema for a single map entry.
 (clojure.core/defrecord MapEntry [key-schema val-schema]
   Schema
   (spec [this]
     (collection/collection-spec
      spec/+no-precondition+
-     vec
+     map-entry-ctor
      [(collection/one-element true key-schema (clojure.core/fn [item-fn e] (item-fn (key e)) e))
       (collection/one-element true val-schema (clojure.core/fn [item-fn e] (item-fn (val e)) nil))]
      (clojure.core/fn [[k] [xk xv] _]
--- a/test/cljx/schema/coerce_test.cljx
+++ b/test/cljx/schema/coerce_test.cljx
@@ -92,3 +92,10 @@
 (deftest constrained-test
   (is (= 1 ((coerce/coercer! (s/constrained s/Int odd?) coerce/string-coercion-matcher) "1")))
   (is (= {1 1} ((coerce/coercer! (s/constrained {s/Int s/Int} #(odd? (count %))) coerce/string-coercion-matcher) {"1" "1"}))))
+
+(deftest map-entry-test
+  (let [entry (first {:foo :bar})
+        coercer (coerce/coercer! (s/map-entry s/Any s/Any) (constantly nil))
+        coerced-value (coercer entry)]
+    (is (= entry coerced-value))
+    (is (= (type entry) (type coerced-value)))))


w01fe:
Thanks for the PR.  In general coercion is not guaranteed to preserve collection type, just to return something that satisfies the schema with the right contents -- and in this case the schema matches anything 'map-entry-like', e.g., which you can conj onto a map.  That said, I could see this as being a bit unexpected in this case.  I'm happy to merge, but for my own edification would you mind explaining the use case where the output type was causing an error for you?  

Thanks again!

imrekoszo:
Well, it's a quirky one and there's probably a better way to do this, but here it is. The data structure I needed to coerce had a few values which sort of acted as demarcation points; the subtrees under those values needed to be coerced according to different rules.

In the following example the ```:asdf``` subtree has different coercion rules, the rest should be coerced via ```fallback-coercer```.

```clojure
(deftest foo-test
  (let [value-aware-coercer (schema.coerce/coercer
                              {:qwer s/Str}
                              (fn [schema]
                                (let [fallback-coercer (schema.coerce/coercer schema (constantly nil))]
                                  (fn [value]
                                    (if (= :asdf value)
                                      :qwer
                                      (fallback-coercer value))))))]
    (is (= {:qwer "asdf"} (value-aware-coercer {:asdf "asdf"})))))
```

This test fails without the fix and the underlying error is a ClassCastException from PersistentVector to java.Map#Entry caught at https://github.com/plumatic/schema/blob/master/src/cljx/schema/coerce.cljx#L40

Does this make any sense?

w01fe:
Thanks for the details.  I can't reproduce the error you're seeing, but I do see a couple issues with the above: 
 - It ends up coercing and validating the data multiple times, since at every level of the data you call a coercer which walks and coerces the whole thing.  This is almost certainly not what you want.  To get the result I think you want, you could just do this:

```clojure
(let [value-aware-coercer (schema.coerce/coercer
                           {:asdf s/Str}
                           (fn [schema]
                             (fn [value] 
                              (if (= "asdf" value)
                                "qwer"
                                value))))]
  (value-aware-coercer {:asdf "asdf"}))
```

 - You probably don't need to coerce all values, but only those that are expected to match a certain schema (this is part of the power of coercers vs walkers).  This could be `s/Str`, or ideally something more specific -- e.g., 

 ```clojure
(let [MySchema (s/named s/Str 'foo)
      value-aware-coercer (schema.coerce/coercer
                           {:asdf MySchema}
                           {MySchema (fn [value] 
                                       (if (= "asdf" value)
                                         "qwer"
                                         value))})]
  (value-aware-coercer {:asdf "asdf"}))
```

Does that make sense? 

 Both of the above work properly on master, but I don't understand what in your version is causing the error you're seeing.  That said, I don't see any harm in merging this, so I'll go ahead and do it.  Thanks again!

**Comment by imrekoszo** [comment]
It does make sense, thank you. I did think about the multiple traversals but figured I'd try to optimize later when I already have a correct result :)

I thought about my usecase a bit more and I'll try to describe it better now. The application is a http api, accepting json among others.

Say the schema is:
```clojure
{:application {:applicant {:name s/Keyword
                           :age  s/Keyword}
               :foo       {:bar s/Keyword}}
 :office      {:address {:street s/Keyword}
               :phone   s/Keyword}}
```

The normal flow is:
json string -> parse into map -> coerce against schema
Works flawlessly.

But there is a different API endpoint where data comes in with a different encoding and after the parse step I have a map like the following:

```clojure
{:application "{\"applicant\":{\"name\":\"foo\",\"age\":\"bar\"},\"foo\":{\"bar\":\"foo\"}}",
 :office "{\"address\":{\"street\":\"bar\"},\"phone\":\"foo\"}"}
```

So when coercion starts, both the :application and :office subtrees are still unparsed strings (json/edn etc).

My extension point is the coersion matcher so I have to write a matcher that parses those values (demarcation points) and then runs the normal json coercion matchers on them.

I would appreciate any further pointers :)

Thanks for the merge!

w01fe:
I see.  Ideally, I would say that you should parse the json outside the
coercion -- it more naturally feels like a different operation to me.  But
if that's not possible in how you set things up, you could use a coercion
matcher like:

(fn [schema]
   (if (instance? clojure.lang.APersistentMap schema)
      (fn [value] (if (string? value) (json/parse value) value)
      (default-coercion-matcher schema)))

This will parse any String that's supposed to be a map schema, then apply
any coercers you have to the innards (assuming you don't have any
additional coercions for the map schemas themselves, only things inside).

Does that sound right?

imrekoszo:
Unfortunately in this case a third party component is tying my hands so I can only provide a matcher. I'm planning to submit a PR there to separate the concerns but until that's done I have to live with the current situation :)

The problem with the approach you described is that I cannot make any assumptions based on the current schema. (the (instance? clojure.lang.APersistentMap schema) part). The schema to be matched could be anything - if the value is detected to be unparsed, it must be parsed and coerced to match the currently active schema. Frustrating, I know :)

w01fe:
Gotcha, that makes sense.  I suppose I can't think of a better solution
than doubly-nested coercion then (besides as you mention fixing the
third-party component), although it's definitely an "off-label" use :)

PR #370, opened 2016-09-12
Fix isa schema for keywords without hierachy

### LINKED ISSUE #369: isa schema not working correctly for keywords without hierarchy
Not sure if I'm using this wrong, but it's not working as I'd expect.

```clojure
(derive :type/Integer :type/Number) 
;; -> nil

(isa? :type/Integer :type/Number)
;; -> true

(s/validate (s/isa :type/Number) :type/Integer)
;; -> ExceptionInfo Value does not match schema: (throws? (isa? :type/Integer :type/Number))  schema.core/validator/fn--8446 (core.clj:155)
```

I can work around it using `pred`, so it's not a deal-breaker.

```clojure
(s/validate (s/pred #(isa? % :type/Number)) :type/Integer) 
;; -> :type/Integer
```

camsaul:
Includes tests. Not sure if these specific changes make sense or if you think things should be organized a bit differently. Happy to change as needed.

Fixes #369

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -266,7 +266,11 @@
 
 (clojure.core/defrecord Isa [h parent]
   Schema
-  (spec [this] (leaf/leaf-spec (spec/precondition this #(isa? h % parent) #(list 'isa? % parent))))
+  (spec [this] (leaf/leaf-spec (spec/precondition this
+                                                  #(if h
+                                                     (isa? h % parent)
+                                                     (isa? % parent))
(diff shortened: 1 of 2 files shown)


w01fe:
This looks great, thanks!  Let me know if it would be helpful to cut a release, otherwise it'll go out whenever I get around to that.

camsaul:
Don't worry about cutting a release on my behalf, I can wait for the next one 👍

w01fe:
OK great, thanks!

PR #383, opened 2017-03-17
Add schema validation error highlight

vpagliari:
Adds _breaklines_ and _yellow color_ on the schema error in order to have a **cleaner and better to read message**;

> Before:
> screen shot 2017-03-17 at 17 28 29

> After:
> screen shot 2017-03-17 at 17 25 43


w01fe:
Looks good, thanks!

PR #388, opened 2017-04-04
update simple-validated-defn-test to account for changes in d4136e630

theJohnnyBrown:


w01fe:
Thanks!  My bad, should have verified that.  Actually, the test exists for a reason; we want the top of the stack to be the user's function, not a helper.  As such, `pretty-exception-info` should either be inlined or be a macro rather than changing the test.  If you want to take a crack at that it would be much appreciated -- thanks again!

w01fe:
Thanks again for the fix -- resolved by reverting in #390 instead, since it wasn't obvious how to fix the issues introduced the right way.

PR #390, opened 2017-04-05
Revert "Pretty print data from schema validation errors"

w01fe:
Reverts plumatic/schema#386

@phillipm This PR introduced reflection and broke the stack traces.  My bad before not running the tests before pushing a new version, I don't have Clojure installed on my main laptop anymore.  Happy to take a version of this that doesn't cause reflection and doesn't pollute the stack traces from s/defn validation errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.1.6
+ * Revert pretty printing of validation errors to remove reflection and fix stack traces.
+ * Add float type to JVM coercers.
+
+## 1.1.5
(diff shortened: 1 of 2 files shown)


w01fe:
Replaces #388

phillipm:
:+1: sorry about that!
I'll look into another approach that avoids the reflection issues

w01fe:
No worries, I should have checked first.  

It's also important that the top of the stack trace from s/defn errors comes from the user's function, so that means any helpers in the main line have to be macros or inlined I think.

PR #395, opened 2017-06-12
README.md: Fixed FancySeq example typo

elimisteve:


elimisteve:
Sorry if I misunderstood the example I'm trying to "correct"! I don't know what the `"s"` does if not specify the particular value of the string that `FancySeq`'s spec is requiring.

w01fe:
The "s" is just a comment/description in the schema (the example is correct).  It should probably be something more explicit I guess, to avoid confusion.  (If the schema was trying to match just "s" it would be `(s/eq "s")` instead. )

elimisteve:
Gotcha. In the other example it had `"k"` as that string, then used `:k` as a key, so I thought those might have to match or something :-).

elimisteve:
My mistake

w01fe:
No worries, I understand how the example could be confusing and appreciate the effort to help.  If you wanted to make it less confusing by changing "s" to "name" and "k" to "category" or something similar we'd gladly take that PR.

PR #396, opened 2017-06-17
added s/ for refd schema ns and fixed example

vladanghene:


w01fe:
Thanks for the PR!  But, these parts of the README look correct to me as-is.  Can you please elaborate on why you think these changes are needed?

vladanghene:
hi @w01fe,
well, I've encountered and used the def macro from schema.core when making schemas so I assumed it was a mising from the readme and wanted to correct that. but since both clojure.core/def and schema.core/def can be used, I guess it's fine like this as you pointed out.

as for the second "fix" of mine, it seemed to be logical to nest the first in the second statement, seeing as the second is nested in the last and I guess the OCD inside me wanted to link all three.

sorry for the bother and mishaps, if any.
cheers

w01fe:
No bother!  Sorry for the confusion. `s/def is not meant for defining schemas, it's for defining values.  e.g. 

```
(def PhoneNumber #"\d\d\d-\d\d\d\d")
(s/def my-phone :- PhoneNumber "555-5555")
```

That said, you can use it without a schema in place of ordinary `def` I guess, but there's no reason to promote that IMO.  Similarly schema does have `s/defschema` for defining schemas but it has some caveats and ordinary `def` is what I would recommend for this case for most users..  

For the other fix, that definitely makes sense -- the problem is that those examples are meant to correspond to the schemas and data above and below this section and changing them breaks the correspondence.  It might make sense to add a newline between the two separate bits if that would make things clearer. 

Best, Jason

frenchy64:
Closing to avoid anyone being misled about `s/def`'s semantics.

PR #399, opened 2017-09-28
Use fully qualified keyword

bfabry:
The :: syntax is used for autoresolving things in the current namespace, ie aliases, it doesn't work with fully qualified names as those aren't in the current ns.

```
Clojure 1.9.0-beta1
Java HotSpot(TM) 64-Bit Server VM 1.8.0_131-b11
        Exit: Control+D or (exit) or (quit)
    Commands: (user/help)
        Docs: (doc function-name-here)
              (find-doc "part-of-name-here")
Find by Name: (find-name "part-of-name-here")
      Source: (source function-name-here)
     Javadoc: (javadoc java-object-or-class-here)
    Examples from clojuredocs.org: [clojuredocs or cdoc]
              (user/clojuredocs name-here)
              (user/clojuredocs "ns-here" "name-here")
boot.user=> (require '[clojure.core :as core])
nil
boot.user=> ::clojure.core/foo

clojure.lang.LispReader$ReaderException: java.lang.RuntimeException: Invalid token: ::clojure.core/foo
             java.lang.RuntimeException: Invalid token: ::clojure.core/foo
boot.user=> :clojure.core/foo
:clojure.core/foo
boot.user=> ::core/foo
:clojure.core/foo
```

### DIFF
--- a/src/clj/schema/experimental/generators.clj
+++ b/src/clj/schema/experimental/generators.clj
@@ -39,12 +39,12 @@
 (defn element-generator [e params]
   (if (vector? e)
     (case (first e)
-      ::schema.spec.collection/optional
+      :schema.spec.collection/optional
       (generators/one-of
        [(generators/return nil)
         (elements-generator (next e) params)])
 
-      ::schema.spec.collection/remaining
+      :schema.spec.collection/remaining
       (do (macros/assert! (= 2 (count e)) "remaining can have only one schema.")
           (generators/vector (sub-generator (second e) params))))
     (generators/fmap vector (sub-generator e params))))


w01fe:
Thanks!  (side note: those namespaces are deprecated, if you're using them you should switch to the https://github.com/plumatic/schema-generators/ project instead.  )

w01fe:
(Does this issue only happen if you load those experimental namespaces, or from the core schema library?   I will also cut a release as soon as I can.)

bfabry:
@w01fe I didn't actually check, we do have code that loads experimental.generators though and I'd expect it only happens if you use those

w01fe:
OK great, thanks!

puredanger:
While things like `::clojure.core/foo` have worked in prior release, that has never been documented behavior, and in 1.9 we have tightened the reader to throw an error on this invalid behavior. The qualifier in an autoresolved keyword must be an alias in the current namespace.

If you really need `:clojure.core/foo`, just say `:clojure.core/foo`, not `::clojure.core/foo`.

w01fe:
Just pushed up a 1.1.7 release with this change, thanks again for the fix.  @puredanger got it, thanks for the explanation.

PR #401, opened 2018-03-12
Error message for multimethods should include the function name

gfredericks:
It will be a derivative gensym in this case, but that's probably the
best that can be done easily. Otherwise it's a completely generic gensym
like `G__4872`, which makes debugging difficult.


w01fe:
Thanks!  I don't release very often these days so just LMK if you would like one.

gfredericks:
I do want to use this change, so if you don't make a release I'll be making a temporary one from a different group.

Which is a good reason to try to start using tools.deps now that I think about it.

PR #402, opened 2018-03-31
Fix for defmethod across aliased ns

aiba:
I think with #401 we lost the ability to defmethod an aliased symbol, e.g.:

```clojure
(require '[foo.bar :as bar])
(s/defmethod bar/some-method ...)
```

Because `(gensym 'bar/some-method)` creates a qualified symbol that causes clojure.core/defmethod to barf.

I think this PR fixes it.


gfredericks:
Yep, that sounds right, thanks!

gfredericks:
I'll make another release for this in the next couple days.

gfredericks:
This is released in `1.1.9`.

PR #405, opened 2018-06-11
Use goog.object/get in place of aget etc (fixes #404).

### LINKED ISSUE #404: Use goog.object/get/set instead of aget/aset
- [x] This is a bug report (with instructions to reproduce) or other issue with the code
(if this is a question or feature request, please **do not** open an issue and post on the [mailing list](https://groups.google.com/forum/#!forum/prismatic-plumbing) instead).  

ClojureScript now warns on the use of `aget`/`aset` with non-integer indexes. I think this code should move to using `goog.object/get` and `set` instead?

https://github.com/plumatic/schema/blob/1b968f95b94f92c0124a4e7f45472417215fcd21/src/cljx/schema/utils.cljx#L157-L161

More info: https://clojurescript.org/news/2017-07-14-checked-array-access

w01fe:
I'm out of touch with the cljs side of things but https://clojurescript.org/news/2017-07-14-checked-array-access suggests this is the right thing to do (HT @danielcompton).

### DIFF
--- a/src/cljx/schema/utils.cljx
+++ b/src/cljx/schema/utils.cljx
@@ -155,10 +155,10 @@
 #+cljs
 (do
   (defn declare-class-schema! [klass schema]
-    (aset klass "schema$utils$schema" schema))
+    (goog.object/set klass "schema$utils$schema" schema))
 
   (defn class-schema [klass]
-    (aget klass "schema$utils$schema")))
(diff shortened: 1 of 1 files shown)


danielcompton:
Looks right to me, but you'll want to add a `goog.object` `:require` as well.

w01fe:
Ah yes, thanks.

PR #407, opened 2018-07-27
Docs: Add license

dijonkitchen:


dijonkitchen:
Frustrating, but changed to V1...

w01fe:
(merging this for now but happy to keep discussing -- perhaps bring it up on the mailing list if you would like to discuss further?)

dijonkitchen:
If the stance is to stay consistent with `clojure`, it'll have to be resolved there first. Seems like `leiningen` is on it too, so I'd expected a wide-sweeping change all at once down the dependencies, if it ever happens. This is a good start.

PR #410, opened 2019-01-17
Use `uuid` in CLJS to avoid warnings

wilkerlucio:
Compiling latest Schema with latest CLJS (1.10.439) + Shadow CLJS latest (2.7.15) generates the warning:

```
------ WARNING #1 --------------------------------------------------------------
 File: schema/coerce.cljs:108:20
--------------------------------------------------------------------------------
 105 |
 106 |
 107 |
 108 |   #(if (string? %) (cljs.core.UUID. %) %))
--------------------------^-----------------------------------------------------
 Wrong number of args (1) passed to cljs.core.UUID
--------------------------------------------------------------------------------
 109 |
 110 |
 111 | (def ^:no-doc +json-coercions+
 112 |   (merge
--------------------------------------------------------------------------------
```

This PR uses the simple `uuid` fn from cljs core that fixes the warning.

### DIFF
--- a/src/cljx/schema/coerce.cljx
+++ b/src/cljx/schema/coerce.cljx
@@ -105,7 +105,7 @@
   #+clj
   (safe #(java.util.UUID/fromString ^String %))
   #+cljs
-  #(if (string? %) (cljs.core.UUID. %) %))
+  #(if (string? %) (uuid %) %))
 
 
 (def ^:no-doc +json-coercions+


w01fe:
Thanks!  Do you know how long the `uuid` function has been in ClojureScript?  Don't want to break backward compatibility if it's a very new addition.  (i tried to look at the blame on github but it broke because `core.cljs` is so big).

wilkerlucio:
@w01fe not so sure when got in, but checking commits from 2017 it was already there: https://github.com/clojure/clojurescript/blob/7cbbff5b0150b13f1134074a2e51fd1467225903/src/main/cljs/cljs/core.cljs#L10536

w01fe:
that's good enough for me, thanks again!

w01fe:
(I can cut a release if you want, just let me know.  I'm not actively using Clojure(Script) myself at the moment (unfortunately), so it might take me a little bit to do otherwise.  If you are interested in being more involved in the project too just LMK)

wilkerlucio:
@w01fe if you can do it I appreciate, so I can get a warning-free compilation, schema is last missing piece :)

w01fe:
Sure thing.  I just cut schema-1.1.10, please let me know if you run into any issues.

PR #412, opened 2019-06-03
Don't catch Throwable on the JVM

gfredericks:
I don't think catching `Throwable` is a good thing for schema to be doing.

This came up when I was running a repl thread that was doing lots of schema completer work, and I couldn't `(.stop thread)` it. Presumably schema was catching the `ThreadDeath` exception and turning it into a completer error.

I think that kind of thing is indicative of the difference between catching & swallowing `Exception` vs `Throwable`.

### DIFF
--- a/src/clj/schema/macros.clj
+++ b/src/clj/schema/macros.clj
@@ -28,7 +28,7 @@
     (assert (symbol? sym))
     `(if-cljs
       (try ~@try-body (~'catch js/Object ~sym ~@catch-body))
-      (try ~@try-body (~'catch Throwable ~sym ~@catch-body)))))
+      (try ~@try-body (~'catch Exception ~sym ~@catch-body)))))
 
 (defmacro error!
   "Generate a cross-platform exception appropriate to the macroexpansion context"


gfredericks:
Another good example of this is an out of memory exception, or a stack overflow; I wouldn't want schema catching those.

gfredericks:
I suspect most use cases that this was aimed at should be instances of `Exception`, but it's hard to be confident that _all_ of them are. So that's the risk of making this change.

w01fe:
Yeah, this is a dirty but Exception is also often not broad enough (e.g. AssertionError is not an Exception).  The right thing is probably something like Scala's `NonFatal` https://github.com/scala/scala/blob/2.13.x/src/library/scala/util/control/NonFatal.scala#L42 but I don't know if there is a similar thing in Clojure land?  I guess we could just copy those cases here... thoughts?

gfredericks:
I didn't know about `NonFatal` -- yes, I think that approach is totally doable, and also conservative w.r.t. backwards-compatibility in that it keeps the present behavior by default.

(I'm not confident that this would have been the best approach from the beginning, but backwards compatibility is important, and this approach solves my problem, so I'm happy with it)

I'll make that change and then merge & release sometime tomorrow probably.

w01fe:
Yeah, I'm not sure what the best thing overall is if we were starting over. But agree that given where we're at, the `NonFatal` approach should be a strict improvement.  Thanks!

gfredericks:
Closing this because #413

PR #413, opened 2019-06-11
Don't swallow fatal exceptions

gfredericks:
Supercedes #412

Some JVM exception types are too severe to be caught and interpreted
as schema failures. This commit adds a short whitelist of those types
and re-throws them.

### DIFF
--- a/src/clj/schema/macros.clj
+++ b/src/clj/schema/macros.clj
@@ -19,8 +19,10 @@
   (if (cljs-env? &env) then else))
 
 (defmacro try-catchall
-  "A cross-platform variant of try-catch that catches all exceptions.
-   Does not (yet) support finally, and does not need or want an exception class."
+  "A cross-platform variant of try-catch that catches all* exceptions.
+   Does not (yet) support finally, and does not need or want an exception class.
+
+   *On the JVM certain fatal exceptions are not caught."
   [& body]
   (let [try-body (butlast body)
         [catch sym & catch-body :as catch-form] (last body)]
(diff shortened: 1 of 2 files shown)


gfredericks:
@w01fe do you have something setup to run the tests? I can't remember what it takes. `lein test` prints a claim of failure, seemingly in the middle of the test run.

gfredericks:
(I was at least able to confirm that one of the tests I added failed in the old code and now passes)

w01fe:
I just ran with `lein test`.   It seems like on master `string-coercer-test` is failing in cljs with "#TypeError: undefined is not an object (evaluating 'schema.coerce.uuid.call)".   Is that what you're seeing?   Not sure when that appeared.  (I'm very sad I never got CI working on this project, I recall some problem with the cljs stuff at the time.)

gfredericks:
I'll try to get CI working; I have it setup for at least one other project w/ cljs, I think, probably maybe.

gfredericks:
I don't think I have permissions to create the build on circleci; can you?

it says to contact the project admin, so I assume I'm not one of those, but maybe I'm wrong.

gfredericks:
I don't think, in the course of developing #414, that I ever saw a failure like `#TypeError: undefined is not an object (evaluating 'schema.coerce.uuid.call)`.

In any case, using the #414 code, this code now passes (I had to make the tiny change in d44b38e to get cljs passing). So if you don't have any comments on the substance, I will merge this once #414 is merged.

w01fe:
Sounds good, thanks a ton for doing this!  I think I've set up circle for schema and schema-generators, and I also made you an admin for the org so hopefully you have permissions to tweak htings.  If not let me know and I'll try harder.

gfredericks:
woah look, it built the branch too!

PR #414, opened 2019-06-12
Modernize dev/test setup

gfredericks:
- Upgrade some deps, most notably cljs
  - also changed clojure versions from 1.7,1.8,1.9 to 1.8,1.9,1.10,
    which I believe was necessary due to some other dependency
  - fix some tests that failed due to changes in cljs
    - most notably, the `fn-name` function has been significantly
      changed, and includes a hard-coded check to return
      `"function"` because that's what the test expects; I don't
      *think* there are edge cases there where the string
      `Function` appears some other way, but it's hard to be sure
- switch from clojurescript.test to doo
  - requires adding a test runner file, which is
    the only cljs-only file, so that required adding
    path entries to a couple places
- modernize the cljsbuild config (and modify it to target
  doo)
  - this includes targeting nodejs instead of phantomjs; my
    impression is that phantomjs is no longer maintained, and
    I think node works fine for these purposes
- add circleci config; the code in this commit is currently passing
  on my fork configuration

Things I didn't do, but could:

- attempt to figure out the minimum supported version of cljs
- add more aliases to test more versions of cljs (I upgraded to
  the latest stable version)
- convert to cljc (I have no idea what difficulties there might
  be for this, and I don't think it's a big deal at the moment)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,41 @@
+version: 2.0
+
+jobs:
+  build:
+    docker:
+      - image: circleci/clojure:lein-2.8.1
+    working_directory: ~/schema
+
+    steps:
+      - checkout
+
+      - restore_cache:
+          # always restore the latest cache
+          key: v1-jars-
+
+      - run: lein deps
+
+      - save_cache:
+          key: v1-jars-{{ checksum "project.clj" }}
+          paths:
+            - ~/.m2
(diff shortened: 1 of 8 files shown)


gfredericks:
Will do.

gfredericks:
@w01fe can you setup the circleci build?

PR #415, opened 2019-06-13
Restrict extra-keys schema in the map spec

gfredericks:
This will fix [issue #16 in schema-generators](https://github.com/plumatic/schema-generators/issues/16),
by wrapping any extra-keys schema in maps in a
s/conditional that asserts that none of the extra
keys can be equal to any explicit key in the map.

This arguably reflects the real meaning of the
extra-key entries, and in any case solves the
schema-generators problem (when generated extra
keys collide with explicit keys) and doesn't
seem to cause any problems.


gfredericks:
This was already discussed [here](https://github.com/plumatic/schema-generators/issues/16#issuecomment-501504500), so I will just merge this when the build passes.

Issue #416, opened 2019-07-03
The CLJS fn-name utility fails if the function name ends in $


zmthy:
Our entire client test suite crashes when upgrading its schema dependency to `1.1.11`. The cause of the crash is this change:

https://github.com/plumatic/schema/commit/2e72125c4b2d8dac3bbed5bd0ac5592e0dce3ec5#diff-023920e7404f0106d749abcaf369621eR66

An advanced optimization by the Closure compiler was adding `$` to either side of the munged name of a function being passed to `pred`, causing the resulting `demunge` to end in a slash. The regex `[^/]+$` that follows fails to match the string, because there is not at least one non-slash character before the end, so `fn-name` returns `nil`. This causes `pred` to throw because `nil` cannot be turned into a symbol.

gfredericks:
Thanks for the detailed report!

Is this easy to write a unit test for?

zmthy:
Yes, since `$` is a valid character in a function name.

Looks like I wasn't quite accurate: it fails if the function ends in at least two `$` characters, since `demunge` will strip a trailing `$` but not all trailing `$`.

Something like this will work:

```clojure
#+cljs (is (not= nil (utils/fn-name (fn foo$$ [x] (+ x x)))))
```

It would be better to test what the actual expected output is, but I'm not sure what that should be: `foo` or `function`?

gfredericks:
would `foo$$` be bad? Perhaps you're thinking correctly returning `foo$$` is impossible because of ambiguity?

zmthy:
I don't think it's ambiguous because there's no way there could have been slashes at the end of the original name, but it depends how consistent you want to be with a function name like `foo$bar`, which is ambiguous. Stripping the trailing `$`s would be ideal for my use-case, because it would undo the rewriting that Closure did, but that doesn't seem like the most general solution. Adding any trailing `$`s on the original name back to the end of the demunged name before it goes into the final regex is probably the right solution (and that would handle the case of the name only consisting of `$`s as well).

gfredericks:
Take a look at #420 and let me know if it addresses your concerns. If not, some additional test cases would be helpful.

zmthy:
Works for me!

gfredericks:
Released as `1.1.12`.

PR #418, opened 2019-07-29
Fix "Wrong number of args" warning in cljs.

chadharrington:
This PR fixes a "Wrong number of args" warning in cljs. You can see the warning in the attached screenshot (shadow-cljs output):

Screen Shot 2019-07-29 at 5 39 56 PM

Specifically, the fix calls the MapEntry constructor with the required third __hash arg. The cljs docs specify that the constructor takes 3 arguments: https://cljs.github.io/api/cljs.core/MapEntry

Merging this tiny PR will make our build output a lot less noisy!

Thank you for your good work on this project.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -753,7 +753,7 @@
 
 (clojure.core/defn map-entry-ctor [[k v :as coll]]
   #+clj (clojure.lang.MapEntry. k v)
-  #+cljs (cljs.core.MapEntry. k v))
+  #+cljs (cljs.core.MapEntry. k v nil))
 
 ;; A schema for a single map entry.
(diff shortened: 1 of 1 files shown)


mynomoto:
I got the same warning and was cloning the repo to do a PR when I found this. This would be really nice for cljs users.

gfredericks:
Looks good, thanks!

gfredericks:
I'd like to fix #416 and then release both of these together, but I can release just this if I don't get around to the other one.

chadharrington:
Thanks, Gary!

w01fe:
👍

gfredericks:
Released as `1.1.12`.

chadharrington:
Thanks!

Issue #417, opened 2019-07-26
Predicate name wrong in CLJS production build


Jarzka:
I create the following predicate
`(schema/pred number?)`

When I print the predicate, I get the following correct output in Clojure and ClojureScript development build using Figwheel:
`#schema.core.Predicate{:p? #object[cljs$core$number_QMARK_], :pred-name number?}`

However, when I create the same predicate in ClojureScript production build with advanced optimizations turned on, I get the following incorrect output.
`#schema.core.Predicate{:p? #object[Xa], :pred-name Xa}]`

This output causes errors in my code since I need to have the correct `:pred-name` value from the predicate. In development build, I get the correct answer (`number?`), but in production build, I get the incorrect answer (`Xa` or some other random string).

Why does the predicate name change in production build? Is there anything we can do to prevent it from happening?

Versions used:
`[prismatic/schema "1.1.11"]` (also happened in `[prismatic/schema "1.1.9"]`)
`[lein-cljsbuild "1.1.7"]`
`[org.clojure/clojurescript "1.10.520"]`

EDIT:

It seems that if I provide the name as a parameter, it works in prod build too:
`(schema/pred number? 'onx-taa-niinq-numero?)`
=> `#schema.core.Predicate{:p? #object[Xa], :pred-name onx-taa-niinq-numero?}`

Also, I noticed that if I call the name creation function manually (the same that is used if I do not provide the name as a parameter), it gives me wrong output in prod build:
`(symbol (schema.utils/fn-name number?))`=> `Xa`
and correct output in dev-build
`(symbol (schema.utils/fn-name number?))`=> `number?`

So the problem happens somewhere in `fn-name` function, I think.

gfredericks:
It sounds like something caused by advanced compilation. I think you can add `^:export` or maybe some other sort of metadata to individual predicates to keep their names from getting munged. If you're not willing to do that, I'm not sure this can be solved.

You could write your own macro that uses the name from the macro call, and just expands to `(schema/pred ~pred '~pred)`.

oliyh:
I am getting this failure on the built-in Str schema: https://github.com/plumatic/schema/blob/master/src/cljx/schema/core.cljx#L371

Also a few of the other built-ins make the same call too - `Int`, `Keyword` and `Symbol` at least I notice. I don't get failures for them - I'm assuming it's blowing up on the first one it comes across.

Thanks

gfredericks:
I believe this is fixed now.

gfredericks:
[...]

PR #419, opened 2019-08-02
Fixing advanced cljs compilation #417

oliyh:
Hi,

This fix worked for me in relation to #417 

Thanks,
Oliy

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -368,7 +368,7 @@
 (def Str
   "Satisfied only by String.
    Is (pred string?) and not js/String in cljs because of keywords."
-  #+clj java.lang.String #+cljs (pred string?))
+  #+clj java.lang.String #+cljs (pred string? 'string?))
 
 (def Bool
   "Boolean true or false"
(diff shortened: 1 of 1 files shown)


gfredericks:
@w01fe this seems fine to me, do you have any objections?

w01fe:
Yeah, seems fine if a bit unfortunate.  

I don't know enough about advanced compilation, is this going to be a problem in other places?  Maybe `pred` if passed a symbol should automatically capture the name or something?

gfredericks:
I think it would be a problem in user code as well.

Automatically capturing the symbol would require converting `pred` into a macro, which would be a technically-breaking change, in that any weirdo doing higher-order stuff with it would have to refactor.

We could add a `(defmacro pred* [p] `(pred ~p '~p))`, and then people would have the option of using that. But that's complicated, requires explaining about why it's even necessary, etc..

w01fe:
Yeah I agree the ship has sailed unfortunately. Thanks again!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gfredericks:
Also the macro approach would have edge cases like `(let [unhelpful-name integer?] (s/pred unhelpful-name))`.

gfredericks:
Released as `1.1.12`.

PR #428, opened 2021-10-27
Proposal: allow maps to have s/precondition-key for use in s/cond-pre

shaunlebron:
## Background
My team uses `s/either` for maps and noticed it was deprecated. I studied the repo to find out why (and in the process I added a [Schema Quick Ref](https://github.com/plumatic/schema/wiki/Schema-Quick-Reference) to the wiki for newcomers).

I learned that `s/cond-pre` relies on hidden preconditions belonging to all core schemas, so maps can’t be used there since they all share the same precondition `map?`.

## Proposal
Add `s/precondition-key` for use in map specs.  Each precondition-key found in a map is appended as a precondition on the map schema.  They are also treated as required keys.

## Example
The wiki mentions a [Shape schema](https://github.com/plumatic/schema/wiki/Defining-New-Schema-Types-1.0#variant-schemas), which we can rewrite with this proposal:

Before:

```clj
(s/defschema Shape
  "A schema for shapes: squares, rectangles, and circles"
  (s/conditional
    #(= (:type %) :square) {:type (s/eq :square) :side s/Num}
    #(= (:type %) :rectangle) {:type (s/eq :rectangle) :width s/Num :height s/Num}
    #(= (:type %) :circle) {:type (s/eq :circle) :radius s/Num}))
```

After:

```clj
(s/defschema Shape
  "A schema for shapes: squares, rectangles, and circles"
  (s/cond-pre
    {(s/precondition-key :type) (s/eq :square) :side s/Num}
    {(s/precondition-key :type) (s/eq :rectangle) :width s/Num :height s/Num}
    {(s/precondition-key :type) (s/eq :circle) :radius s/Num}))
```

## Tests

I added basic tests for `s/precondition-key` and added more cases to `s/cond-pre` that use it.

## Blind Spots

I think I understand that `s/either` was deprecated for better efficiency and errors. And I have a general sense that a tagged union should have a proper way to compute its claimed type (tag) without validating everything in it.  But I don’t understand why it breaks coercion.  So this proposal comes with the caveat that I may be missing the purpose of deprecating it.

### DIFF
--- a/src/cljx/schema/core.cljx
+++ b/src/cljx/schema/core.cljx
@@ -703,6 +703,7 @@
 ;; Specific keys are mapped to value schemas, and given as either:
 ;;  - (required-key k), a required key (= k)
 ;;  - a keyword, also a required key
+;;  - (precondition-key k), a required key used as a precondition
 ;;  - (optional-key k), an optional key (= k)
 ;; For example, {:a Int (optional-key :b) String} describes a map with key :a mapping to an
 ;; integer, an optional key :b mapping to a String, and no other keys.
@@ -726,6 +727,16 @@
     k
     (RequiredKey. k)))
 
+(clojure.core/defrecord PreconditionKey [k])
+
+(clojure.core/defn precondition-key
+  "A required key in a map used as a precondition"
+  [k]
+  (PreconditionKey. k))
+
+(clojure.core/defn precondition-key? [ks]
+  (instance? PreconditionKey ks))
+
 (clojure.core/defn required-key? [ks]
   (or (keyword? ks)
       (instance? RequiredKey ks)))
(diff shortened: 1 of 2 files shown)


w01fe:
Thanks for writing this up! 

I agree with @frenchy64 that it would be preferable to build something on top of the existing `conditional` support, unless there is something I'm missing that is gained by adding this additional complexity into the core of the library.  Can you say more about why you chose this approach? 

As for your question about coercion, the problem with coercion is that you need to be able to figure out what kind of thing something is supposed to be in order to coerce it to that thing.  With `either` you would have to attempt coercion on each branch and take the first which succeeds; this might be okay in some cases, but seems like it could produce unexpected behavior or performance problems in others.

frenchy64:
TIL about [schema-refined.core/dispatch-on](https://github.com/KitApps/schema-refined#sum-types) which seems more general than my example.

shaunlebron:
> the problem with coercion is that you need to be able to figure out what kind of thing something is supposed to be in order to coerce it to that thing. With `either` you would have to attempt coercion on each branch and take the first which succeeds; this might be okay in some cases, but seems like it could produce unexpected behavior or performance problems in others.

Thanks @w01fe, this is very clear now 👍

> TIL about [schema-refined.core/dispatch-on](https://github.com/KitApps/schema-refined/tree/711e9dcc240f6931be1f6697a891cb28231ca778#sum-types)

Thanks for the reference!  That’ll do.  Closing now 👋

PR #429, opened 2021-11-03
Support mutually recursive `s/letfn`

frenchy64:
I noticed `s/letfn` is unlike `letfn` in that it does not support mutually recursive bindings.

### DIFF
--- a/src/cljc/schema/core.cljc
+++ b/src/cljc/schema/core.cljc
@@ -1396,30 +1396,33 @@
 
 #?(:clj
 (defmacro letfn
-  "s/letfn : s/fn :: clojure.core/letfn : clojure.core/fn"
+  "s/letfn : s/fn :: clojure.core/letfn : clojure.core/fn
+  
+  Gotchas:
+  - s/fn-schema will only work on direct usages of the bindings
+    in the body. It will not work on intermediate calls between bindings."
   [fnspecs & body]
   (let [{:keys [outer-bindings
                 fnspecs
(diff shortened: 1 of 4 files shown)


src/cljc/schema/core.cljc
```diff
@@ -1398,10 +1398,28 @@
...
+                                                               ~schema-form)))))
+                {:outer-bindings []
+                 :fnspecs []
+                 :declare-class-schemas []}
+                fnspecs)]
+    `(let ~outer-bindings
+       (clojure.core/letfn
+         ~fnspecs
+         (do
+           ~@declare-class-schemas
```
w01fe, line 1421: My memory is a little rusty, but I think this is mutating some global variable, which seems weird.    I guess you get what you ask for if you put a `defn` or `defrecord` inside a `letfn` though, so this is probably the least surprising option?

frenchy64: Yes, it's mutating a weak hash map. I went back and forth about this, since the `fn-schema` cannot be seen in the bindings if you use metadata.

eg.,
```
(s/fn-schema (letfn [(foo [] foo)] (foo)))
```

However one of my assumptions I just realized was wrong: a fn can't see it's own metadata.

```
user=> ((with-meta (fn this [] (meta this)) {:a 1}))
nil
```

So `s/fn` has the same problem. I think the metadata impl makes more sense to me now.

w01fe (approved):
Nice, thanks!

PR #430, opened 2021-11-05
Fix `s/fn` performance issues in Clojure

frenchy64:
While trying to implement `s/defprotocol`, I needed the following invariant to hold in Clojure to access [__methodImplCache](https://github.com/clojure/clojure/blob/b8132f92f3c3862aa6cdd8a72e4e74802a63f673/src/clj/clojure/core_deftype.clj#L609):

```clojure
(let [f (s/fn this [] this)]
  (identical? f (f)))
```

The fix I chose was to use `^{:schema schema} (fn ...)` instead of `(with-meta (fn ...) {:schema schema})`. I believe this also fixed all the documented performance caveats for `s/fn` in Clojure as it means wrappers are eliminated.

The perf characteristics for CLJS seem unchanged (still uses the wrapper). I worked around [CLJS-968](https://clojure.atlassian.net/browse/CLJS-968) by placing the function evaluation in non-tail position. I decided working around this issue was better than using `macros/if-cljs` (which I treat as a last resort).

I also moved a test, please ignore whitespace changes when reviewing.

### DIFF
--- a/src/cljc/schema/core.cljc
+++ b/src/cljc/schema/core.cljc
@@ -1276,20 +1276,27 @@
       don't supply a name, schema will gensym one for you and attach
       the schema.
     - Unlike s/defn, the function schema is stored in metadata on the
-      fn.  Clojure's implementation for metadata on fns currently
-      produces a wrapper fn, which will decrease performance and
-      negate the benefits of primitive type hints compared to
-      clojure.core/fn."
+      fn. The implications of this differ per platform:
+      :clj   The resulting function has the same performance characteristics
+             as clojure.core/fn. Additionally, the following invariant
+             holds for all parameters and schema annotations:
+               (let [f (s/fn this ... [...] this)]
+                 (assert (identical? f (f ...))))
+      :cljs  Returns a wrapper function that forwards arguments positionally
+             up to 20 arguments, and then via `apply` beyond 20 arguments.
+             See `cljs.core/with-meta` and `cljs.core.MetaFn`."
   [& fn-args]
   (let [fn-args (if (symbol? (first fn-args))
                   fn-args
                   (cons (gensym "fn") fn-args))
         [name more-fn-args] (macros/extract-arrow-schematized-element &env fn-args)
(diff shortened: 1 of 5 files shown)


src/cljc/schema/core.cljc
```diff
@@ -1276,20 +1276,27 @@
...
+             up to 20 arguments, and then via `apply` beyond 20 arguments.
+             See `cljs.core/with-meta` and `cljs.core.MetaFn`."
   [& fn-args]
   (let [fn-args (if (symbol? (first fn-args))
                   fn-args
                   (cons (gensym "fn") fn-args))
         [name more-fn-args] (macros/extract-arrow-schematized-element &env fn-args)
         {:keys [outer-bindings schema-form fn-body]} (macros/process-fn- &env name more-fn-args)]
-    `(let ~outer-bindings
-       (schematize-fn
```
w01fe, line 1290: Can schematize-fn be deleted/deprecated now?

w01fe (approved):
Very nice, thanks!

src/cljc/schema/core.cljc
frenchy64: Looks like it's used by the most recent changes to `s/letfn` :)

frenchy64: Hmm, with the knowledge gained here, now the tradeoffs are less obvious about whether to use schematize-fn or declare-class-schema! in `s/letfn`...

frenchy64: Ah, perhaps we can have our cake and eat it too by using the same trick in `s/letfn`:

```clojure
(s/letfn [(f [a ...])]
  ...)
=>
(letfn [(f [a ...])]
  (let [f ^{:schema ...} (fn [a ..]))]
    ...))

frenchy64: Created an issue to track this https://github.com/plumatic/schema/issues/431

frenchy64:
There was an oversight in `lein all` that meant the latest cljs was never tested. I needed to drop Clojure 1.7 from the test suite because CLJS doesn't work with it (but we only support Clojure 1.8 and later anyway).

Issue #356, opened 2016-04-27
Validation in ClojureScript extremely slow


sickill:
I have defined schema'd record (here https://github.com/asciinema/asciinema-player/blob/19c5c423e3ca78df1ae04dcebbf0d32437e303e1/src/asciinema/player/vt.cljc#L56) and enabled schema validation in test via `(use-fixtures :once schema.test/validate-schemas)` (here https://github.com/asciinema/asciinema-player/blob/19c5c423e3ca78df1ae04dcebbf0d32437e303e1/test/asciinema/player/vt_test.cljc#L15).

I have noticed that schema validation is very slow when under ClojureScript. Without validation the tests under both Clojure and ClojureScript were finishing in few seconds. After enabling the validation I noticed that it's still couple of seconds for Clojure but **15+ minutes** for ClojureScript (phantomjs). Right now I'm enabling test validation for Clojure only because of this, using reader conditional (as seen in the second link above).

It seems it's not test-specific problem, because I can observe the same behavior when using `(s/set-fn-validation! true)`.

I tried to find a specific part of this `VT` schema that makes it so slow (changing different sub-schemas to `s/Any` etc) and it seems there's no single key/sub-schema that makes it slow, it just adds up from all sub-schemas.

I'd be happy to give more information, test, dig deeper. Any hints?

w01fe:
Thanks for the report.  Is this possibly the same issue as #352 ?  Are you observing failures, or is it slow despite everything succeeding? 

I don't know much about ClojureScript profiling, but I'll try to take a look and see what I can find.  If you can share a slow js file that I can profile in Chrome, that would be a big help.  Thanks!

sickill:
It's always succeeding so not sure if the underlying issue is related to the mentioned one by you.
I'll prepare slow .js file for Chrome, hold on.

sickill:
I pushed compiled files with simple html file to this repo: https://github.com/sickill/gh-356 
It has been compiled without advanced optimization (so it's easier to see function names in profiler), with `(s/set-fn-validation! true)`.

sickill:
I have added CPU profile dump to the repo, you can load it in Chrome profiler. Lot's of `schema.spec.element_transformer` and `schema.spec.collection.CollectionSpec.schema$spec$core$CoreSpec$checker$arity$2` calls. Stack is very deep there, and it's hard for me to see where it actually ends.

sickill:
[...]

PR #436, opened 2022-02-05
Optimize CollectionSpec checker

nilern:
* Replace a `doseq` with `reduce` on JVM
* Accumulate into a JS array on cljs instead of a persistent vector in an atom

The latter could improve #356 substantially; in the Malli seqex impl [I had a similar situation](https://www.metosin.fi/blog/malli-regex-schemas/#memoization) with a persistent set in an atom on cljs and it was just unacceptably slow.

### DIFF
--- a/src/cljc/schema/spec/collection.cljc
+++ b/src/cljc/schema/spec/collection.cljc
@@ -23,11 +23,10 @@
       (let [_ (macros/assert! (= 2 (count e)) "remaining can have only one schema.")
             c (spec/sub-checker (second e) params)]
         #?(:clj (fn [^java.util.List res x]
-                  (doseq [i x]
-                    (.add res (c i)))
+                  (reduce (fn [res i] (doto res (.add (c i)))) res x)
                   (then res nil))
            :cljs (fn [res x]
-                   (swap! res into (map c x))
+                   (reduce (fn [res i] (doto res (.push (c i)))) res x)
                    (then res nil)))))
 
     (let [parser (:parser e)
@@ -35,7 +34,7 @@
       #?(:clj (fn [^java.util.List res x]
                 (then res (parser (fn [t] (.add res (if (utils/error? t) t (c t)))) x)))
          :cljs (fn [res x]
-                 (then res (parser (fn [t] (swap! res conj (if (utils/error? t) t (c t)))) x)))))))
+                 (then res (parser (fn [t] (.push res (if (utils/error? t) t (c t)))) x)))))))
 
 (defn- sequence-transformer [elts params then]
   (macros/assert! (not-any? #(and (vector? %) (= (first %) ::remaining)) (butlast elts))
@@ -58,7 +57,7 @@
 
 :cljs
 (defn- has-error? [l]
-  (some utils/error? l)))
+  (.some l utils/error?)))
 
 (defn subschemas [elt]
   (if (map? elt)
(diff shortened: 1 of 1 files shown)


src/cljc/schema/spec/collection.cljc
```diff
@@ -23,19 +23,18 @@
       (let [_ (macros/assert! (= 2 (count e)) "remaining can have only one schema.")
             c (spec/sub-checker (second e) params)]
         #?(:clj (fn [^java.util.List res x]
-                  (doseq [i x]
-                    (.add res (c i)))
+                  (reduce (fn [res i] (doto res (.add (c i)))) res x)
```
w01fe, line 26: Why is this expected to be faster?  The `List` can't be used safely concurrently so I would be surprised if you could do better than the `doseq`.  Have you profiled and seen a speedup?

src/cljc/schema/spec/collection.cljc
```diff
@@ -23,19 +23,18 @@
...
                   (then res nil))
            :cljs (fn [res x]
-                   (swap! res into (map c x))
+                   (reduce (fn [res i] (doto res (.push (c i)))) res x)
```
w01fe, line 29: I believe that something mutable could be faster here, but I'm again unsure about the use of `reduce`.  Can you share profiling results?

src/cljc/schema/spec/collection.cljc
```diff
@@ -58,7 +57,7 @@
 
 :cljs
 (defn- has-error? [l]
-  (some utils/error? l)))
+  (.some l utils/error?)))
```
w01fe, line 60: I'm not super familiar with cljs, can you explain the change here please?

w01fe:
Thanks for the PR!  Certainly happy to accept optimizations but always prefer if they come with measurements.  And would like to learn more about the choice of `reduce` over `doseq`.

src/cljc/schema/spec/collection.cljc
nilern: `reduce` (and `run!`) is faster and more memory efficient because as an internal iterator it can be fully specialized to the collection and usually does not need to allocate at all. I have profiled it sometime but by now it is just a habit like using `(into [] (map ...) x)` instead of `(into [] (map ... x))`. Since `reduce` works on all seqs really the only reason to ever use `doseq` is if `:when`/`:while`/`:let` is super useful.

w01fe: The weird thing about `reduce` here is that it's threading through an extra argument that you don't care about because you're mutating the argument anyway.   I think it makes the code slightly less clear and I would be a bit surprised if there was a performance benefit, but I'd be happy to believe it if there were accompanying measurements.

src/cljc/schema/spec/collection.cljc
w01fe: Thanks.  Is that expected to be appreciably faster?

w01fe:
Thanks!  I just want to make sure we are striking the right balance between performance and clarity, i.e. only making optimizations that make a measurable difference.  My intuition is admittedly stale here, so I'm happy to take these with measurements, or happy if another maintainer wants to take them without.

src/cljc/schema/spec/collection.cljc
frenchy64: > The weird thing about reduce here is that it's threading through an extra argument that you don't care about because you're mutating the argument anyway.

Might be worth comparing to `(reduce #(.add res (c %2))) nil x)`.

frenchy64: Oh, it would probably beat the current code handily since I don't see a type hint on the inner `res`.

PR #437, opened 2022-03-29
updated README post 1.1.0

The-Alchemist:
v1.1.0 came out about six years ago (almost to the day!), so I think it's safe to remove the numerous 1.1.0 references.

I also streamlined the list of related libraries a bit


w01fe (approved):
Thanks!  This had been bothering me subconsciously for awhile.

PR #438, opened 2022-04-28
s/constructor/konstructor/ to avoid cljs edge case

williammizuta:
Replace constructor to konstructor due to some issues with cljs

See plumatic/schema-generators#22


w01fe (approved):
Awesome, thanks!

PR #440, opened 2022-05-18
Babashka support

frenchy64:
The next version of [babashka](https://github.com/babashka/babashka) will be able to support schema with these changes.

The main thing to note here is that babashka cannot dynamically create classes (since it runs on graalvm), so `cc/fn`, `cc/defrecord`, `cc/defprotocol` have very different implementations than JVM Clojure.

Here, it mostly amounts to broadening `class?` checks to also account for `sci.lang.Type`, which is babashka's representation of dynamically generated classes.

We also use fn identity for `s/fn-schema` since bb keeps a pool of its own `fn` classes for each arity, and revert to less optimized CLJS approach in some other places.

Note: in bb, both `:clj` and `:bb` reader conditionals are active.

Note: in `schema.macros`, we can't use reader conditionals (`.clj`) so `if-bb` uses system props to determine platform.

Note: bb doesn't support extending `print-method` via a protocol, so `macros/defrecord-schema` is used to extend it for each schema primitive.

Bonus: adds pretty `pr-str` printing in CLJS.

The entire test suite remarkably runs in 0.5 seconds:

```
$ bash <(curl https://raw.githubusercontent.com/babashka/babashka/master/install) --version 0.8.3-SNAPSHOT --dir .
$ time ./bb test

Testing schema.core-test

Testing schema.macros-test

Testing schema.coerce-test

Testing schema.experimental.abstract-map-test

Testing schema.test-test

Testing schema.utils-test

Ran 121 tests containing 660 assertions.
0 failures, 0 errors.
./bb test  0.46s user 0.05s system 98% cpu 0.517 total
```

### DIFF
--- a/src/clj/schema/macros.clj
+++ b/src/clj/schema/macros.clj
@@ -220,8 +220,6 @@
      (or (:always-validate fn-meta)
          *assert*))))
 
-(def ^:dynamic *ufv-sym* 'ufv__) ;; dynamic for backwards compat
-
 (defn process-fn-arity
   "Process a single (bind & body) form, producing an output tag, schema-form,
    and arity-form which has asserts for validation purposes added that are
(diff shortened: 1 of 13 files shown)


src/clj/schema/macros.clj
```diff
@@ -201,13 +220,17 @@
      (or (:always-validate fn-meta)
          *assert*))))
 
+(def ^:dynamic *ufv-sym* 'ufv__) ;; dynamic for backwards compat
```
w01fe, line 223: I think it's worth a docstring since it's not obvious to me what's going on here.  Also maybe worth elaborating on the backwards compat concern.

w01fe (approved):
:cool:

src/clj/schema/macros.clj
frenchy64: Thanks for bringing this up. It's a little more hygienic to use fresh symbols to let-bind variables. eg., `(s/fn [ufv__])` would capture the binding.


In terms of implementation, I didn't want to add an extra arity to `process-fn-arity` because 6 args seemed too much, but I'm considering adding a 1-arg arity that just takes a map. That will obviate the need for this dynamic var.

PR #451, opened 2022-10-01
Proof of concept: `s/defmulti` + `arguments-checker`

frenchy64:
`arguments-checker` implements idea outlined in https://github.com/plumatic/schema/issues/450.

Multimethod wrapping for non-graalVM JVM Clojure only atm.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -32,6 +32,7 @@
   :jar-exclusions [#"\.swp|\.swo|\.DS_Store"]
 
   :source-paths ["src/clj" "src/cljc"]
+  :java-source-paths ["src/java"]
 
   :test-paths ["test/clj" "test/cljc" "test/cljs"]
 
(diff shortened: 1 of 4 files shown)


w01fe:
Interesting, thanks for exploring this direction.  I know this is just a draft for now, not sure if you are looking for feedback on the whole thing, or just the arguments-checker piece, or just putting it up for CI? 

In the future I would ideally suggest making a proposal before putting up an implementation for a new feature like `s/defmulti`, but that's of course up to you.  For historical context, we had deliberately omitted this one because it wasn't clear there was significant value above and beyond what's provided by `s/defmethod`, especially given the necessary complexity to implement it.  

In any case, I'm open to reconsidering that but to be able to usefully comment on something like this I would really appreciate at some point an explanation of what user need you're trying to address, how you went about solving it and new user-facing syntax,  explanation behind tricky implementation details (why did you need to drop to java, how safe is this from a maintenance perspective), etc.  Thanks!

frenchy64:
Someone in `#malli` in clojurians wanted similar functionality for malli and I thought I'd try it in Schema since I'm more comfortable here. I will move this to my repo since this is just exploratory.

PR #458, opened 2025-08-18
Address deprecated build dependencies

frenchy64:
Some of our build deps are deprecated causing builds to fail.

For example, here's an error message from a [build from August 18th 2025](https://github.com/frenchy64/schema/actions/runs/17047655194/job/48327485861):

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
```

I also updated the JVM versions we test against and bumped other outdated Actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,8 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


w01fe (approved):
Thanks!

